PR #192, opened 2026-03-05
feat(retrieval): multi-hook retrieval aliases — closes #149

### LINKED ISSUE #149: Feature: Multi-hook retrieval aliases — generate alternative search phrases per fact
## Summary

When storing a fact, automatically generate 3-5 alternative phrasings ("retrieval hooks") and index them alongside the primary embedding — so the fact can be found from multiple semantic angles, not just the one it was written from.

## Motivation

This is the one genuinely clever idea from Total Recall's "multi-hook retrieval." The problem it solves is real even with vector embeddings:

> "Fitbit auth failure" and "browser cookie issue" might describe the same memory, but the embeddings could be far apart in vector space.

LanceDB's cosine similarity is good, but it's still matching against a single embedding per fact. If the fact was stored as "Fitbit auth keeps failing" but the query is "why won't my fitness tracker sync", the semantic distance might push it below the retrieval threshold.

Total Recall solves this with text aliases. We can do better: **store multiple embeddings per fact.**

### At storage time (`memory_store`)
1. After storing the primary fact text and its embedding, call a cheap model to generate 3-5 alternative phrasings:
   - Different vocabulary ("auth failure" → "login problem", "credential issue")
   - Different angles ("Fitbit" → "wearable", "fitness tracker")
   - Causal framing ("cookie expired" → "session timed out")
2. Generate embeddings for each alternative
3. Store in a `fact_aliases` table: `(factId, aliasText, embedding)`

### At retrieval time (`memory_recall`)
- Search across both primary embeddings AND alias embeddings
- Deduplicate results (multiple aliases hitting the same fact → return it once with best score)
- No change to the retrieval API — this is a backend-only improvement

## Cost considerations

- Generation: ~$0.001 per fact (cheap model, short prompt)
- Storage: ~3-5x more embeddings in LanceDB (minimal disk impact)
- Retrieval: slightly more vectors to search, but LanceDB handles this efficiently
- **Only generate aliases for importance ≥ 0.5** to avoid wasting compute on low-value facts

## When NOT to generate aliases

- Session-scoped facts (too transient)
- Facts that are already very short/specific (entity names, dates)
- Low-importance observations

## Source

Total Recall's "multi-hook retrieval" concept — reimplemented properly with actual vector embeddings instead of text string matching.

---

In Design A, this is implemented with a single local embedding model generating aliases. The schema should support N vectors per fact from day one (even if A only uses 1 model + aliases). This enables Design B's multi-model embeddings later without schema changes.

Add a `fact_embeddings` table:
```sql
CREATE TABLE fact_embeddings (
  id TEXT PRIMARY KEY,
  fact_id TEXT NOT NULL,
  model TEXT NOT NULL,       -- 'nomic-embed-v1.5', 'alias', etc.
  variant TEXT,              -- 'canonical', 'alias-1', 'contextual-means', 'contextual-search'
  embedding BLOB NOT NULL,
  created_at TEXT NOT NULL
);
```
This replaces storing embeddings directly in the facts table and supports any number of vectors per fact.

markus-lassfolk:
Closes #149.

---

> [!NOTE]
> **Medium Risk**
> Adds a new optional retrieval strategy and SQLite database plus LLM/embedding work on store paths; although gated behind `cfg.aliases.enabled` (default off), it touches core store/recall/delete flows and could affect performance or cleanup correctness when enabled.
> 
> **Overview**
> Adds an *optional* “retrieval aliases” feature (`cfg.aliases`) that generates up to N alternative phrasings per stored fact, embeds them, and persists them in a new SQLite `AliasDB` (`aliases.db`).
> 
> Integrates alias search into `runRetrievalPipeline` as an additional RRF strategy (`source: "aliases"`) and wires `aliasDb` through plugin/CLI/tool/lifecycle contexts; alias rows are also cleaned up whenever facts are deleted/superseded/merged/migrated (store/update/delete, consolidation, credential migration, diagnostics). Includes a new `retrieval-aliases` service with comprehensive tests.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7b34dc9cc007924644ba40e7e5b577b6210db36f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/extensions/memory-hybrid/cli/handlers.ts
+++ b/extensions/memory-hybrid/cli/handlers.ts
@@ -254,6 +254,7 @@ export interface HandlerContext {
   openai: OpenAI;
   cfg: HybridMemoryConfig;
   credentialsDb: CredentialsDB | null;
+  aliasDb: import("../services/retrieval-aliases.js").AliasDB | null;
   wal: WriteAheadLog | null;
   proposalsDb: ProposalsDB | null;
   resolvedSqlitePath: string;
@@ -293,7 +294,7 @@ export async function runStoreForCli(
   opts: StoreCliOpts,
   log: { warn: (m: string) => void },
 ): Promise<StoreCliResult> {
-  const { factsDb, vectorDb, embeddings, openai, cfg, credentialsDb } = ctx;
+  const { factsDb, vectorDb, embeddings, openai, cfg, credentialsDb, aliasDb } = ctx;
   const text = opts.text;
   if (factsDb.hasDuplicate(text)) return { outcome: "duplicate" };
   const sourceDate = opts.sourceDate ? parseSourceDate(opts.sourceDate) : null;
@@ -395,6 +396,7 @@ export async function runStoreForCli(
           if (classification.action === "NOOP") return { outcome: "noop", reason: classification.reason ?? "" };
           if (classification.action === "DELETE" && classification.targetId) {
             factsDb.supersede(classification.targetId, null);
+            aliasDb?.deleteByFactId(classification.targetId);
             return { outcome: "retracted", targetId: classification.targetId, reason: classification.reason ?? "" };
           }
           if (classification.action === "UPDATE" && classification.targetId) {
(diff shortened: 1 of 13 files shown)


markus-lassfolk:
Superseded by unified merge PR (all remaining features combined into single PR).

Issue #206, opened 2026-03-06
feat: Document ingestion via MarkItDown Python bridge


markus-lassfolk:
## Summary
Add a `memory_ingest_document` tool that converts documents (PDF, DOCX, XLSX, PPTX, images, audio, HTML, etc.) to Markdown via [Microsoft MarkItDown](https://github.com/microsoft/markitdown), then intelligently chunks and stores them as facts with source attribution.

## Architecture
```
memory_ingest_document(path, opts)
  → python-bridge.ts (persistent Python subprocess, JSON-RPC over stdin)
  → chunker.ts (heading-aware markdown splitting)
  → factsDb.store() per chunk (with source_document, page, section metadata)
  → embeddings via existing pipeline
```

## Key Design Decisions
1. **Persistent Python subprocess** — spawned at plugin init, kept alive, reused across calls (~500ms startup saved per call)
2. **Thin Python bridge** (~30 lines) — imports MarkItDown, reads JSON-RPC from stdin, returns markdown + metadata
3. **Chunking in Node.js** — hybrid-memory knows fact sizes, embedding dimensions, token limits
4. **Source tracking** — each chunk gets `source_document`, `page_number`, `section_heading` metadata
5. **Deduplication** — re-ingesting same document updates existing facts (hash-based)
6. **Config flag** — `documents.enabled: true` with `pip install 'markitdown[all]'` as optional dependency

## Features
- [ ] Python bridge service (spawn, health check, restart on crash)
- [ ] `memory_ingest_document(path, opts)` tool
- [ ] Heading-aware markdown chunker with configurable chunk size
- [ ] Source attribution metadata on stored facts
- [ ] Deduplication on re-ingestion (hash-based)
- [ ] Batch ingestion (`memory_ingest_folder`)
- [ ] Progress callbacks for large files
- [ ] LLM vision integration for images using existing embedding provider config
- [ ] Config schema additions (`documents.*`)

## References
- MarkItDown: https://github.com/microsoft/markitdown
- MCP server (ad-hoc use): https://github.com/microsoft/markitdown/tree/main/packages/markitdown-mcp

markus-lassfolk:
## 📋 Gap Audit (PR-aware) — Issue #206

### ✅ Implemented
- (PR #256) memory_ingest_folder tool, progress callbacks, LLM vision integration, vision config
- (PRs #213, #216) Python bridge, memory_ingest_document, chunker, source attribution, deduplication

[...]

Issue #163, opened 2026-03-05
Provenance: Full fact-to-source tracing


markus-lassfolk:
## Summary
Track the complete provenance chain for every fact: from the knowledge store back through consolidation events to the original session and conversation turn where the information was first captured.

## Motivation
When a fact seems wrong, there's currently no way to answer 'why does the system believe this?' or 'where did this come from?' Provenance chains enable:
- Debugging incorrect facts
- Assessing trustworthiness (explicit user statement > passive extraction > inference)
- Auditing the knowledge base
- Re-processing if extraction logic improves

### Provenance edges
Every fact MUST have at least one DERIVED_FROM edge indicating origin:
- `DERIVED_FROM → event_log:event-123` (passively extracted from session event)
- `DERIVED_FROM → active_store:session-456` (explicitly stored by agent in session)
- `DERIVED_FROM → consolidation:[fact-a, fact-b, fact-c]` (merged from multiple sources)

### Additional metadata
```sql
ALTER TABLE facts ADD COLUMN source_session TEXT;
ALTER TABLE facts ADD COLUMN source_turn INTEGER;      -- conversation turn number
ALTER TABLE facts ADD COLUMN extraction_method TEXT;   -- 'active', 'passive', 'consolidated', 'reflected'
ALTER TABLE facts ADD COLUMN extraction_confidence REAL; -- how confident the extraction was
```

### Retention
- Design A: provenance data retained 90 days (same as event log)
- Design B: provenance data retained 1 year
- Configurable via `hybridMemory.provenance.retentionDays`

### Querying
```
memory_provenance(factId) → {
  fact: { id, text, confidence, ... },
  source: { session_id, timestamp, turn, extraction_method },
  derivedFrom: [ { event_id, event_text, timestamp }, ... ],
  consolidationChain: [ ... ]  // if consolidated from earlier facts
}
```

## Priority
P1 within Design B — Important for trust and debugging.

Issue #162, opened 2026-03-05
Storage: Verification Store for critical facts


markus-lassfolk:
## Summary
Add a dedicated high-trust storage tier for critical facts (infrastructure configs, credentials, medical info, financial data) that provides immutability, integrity checking, and protection against silent decay or corruption.

## Motivation
Not all facts are equal. 'Markus prefers dark mode' and 'Proxmox root password is X' have wildly different consequences if forgotten or corrupted. The Verification Store ensures critical facts:
- Can never be silently modified (append-only, versioned)
- Can never auto-decay or auto-archive
- Are integrity-checked on retrieval (corruption detection)
- Have a backup outside the main database

### Schema
```sql
CREATE TABLE verified_facts (
  id TEXT PRIMARY KEY,
  fact_id TEXT NOT NULL REFERENCES facts(id),
  canonical_text TEXT NOT NULL,
  checksum TEXT NOT NULL,          -- SHA-256 of canonical_text
  verified_at TEXT NOT NULL,
  verified_by TEXT NOT NULL,       -- 'agent', 'user', 'system'
  next_verification TEXT,          -- when to re-verify
  version INTEGER DEFAULT 1,
  previous_version_id TEXT,        -- link to superseded version
  created_at TEXT NOT NULL
);
```

### Backup
Verified facts are also written to a flat JSON file (`~/.openclaw/verified-facts.json`) that can survive database corruption. This file is append-only and synced on every write.

### Behavior
- Facts marked as `verification_tier: 'critical'` (manually or by auto-classification of infrastructure/credential facts) are copied to the Verification Store
- On retrieval, checksum is verified against stored text. Mismatch = corruption alert.
- Verified facts are NEVER auto-archived, auto-deleted, or confidence-decayed below 0.5
- Continuous verification (optional): every 30 days, present to LLM: 'Is this still accurate based on recent knowledge?'

### Auto-classification
Facts with these signals are auto-suggested for verification:
- Entity is an IP address, hostname, or credential reference
- Category is 'technical' + tags contain 'infrastructure'
- Explicitly marked by agent or user

### Config
```yaml
hybridMemory:
  verification:
    enabled: false              # Design B feature
    backupPath: '~/.openclaw/verified-facts.json'
    reverificationDays: 30
    autoClassify: true
```

## Priority
P0 within Design B — Directly addresses the 'infrastructure facts unreliable' pain point.

markus-lassfolk:
[...]

Issue #161, opened 2026-03-05
Retrieval: LLM re-ranking of retrieval results


markus-lassfolk:
## Summary
After RRF fusion produces a ranked list, add an optional LLM re-ranking step where a mid-tier model reviews the top-N candidates and re-ranks by semantic relevance to the original query.

## Motivation
Algorithmic ranking (RRF + confidence + recency) is fast and cheap but can't understand nuance. An LLM can understand that 'when did the router IP change' should rank a supersession chain higher than a loosely related networking fact.

Estimated impact: catches 5-10% of cases where algorithmic ranking puts the right fact at position 15 instead of position 3.

### Input
Top-50 facts from RRF fusion, plus the original query.

### Prompt
```
Query: [original query]

Rank these facts by relevance to the query. Return the top 20 as a JSON array of fact IDs, most relevant first.

Facts:
1. [fact-id]: [text] (confidence: X, stored: Y)
2. ...
```

### Cost
- ~2000-3000 tokens per re-ranking call (mid-tier model)
- At 30 retrievals/day: ~$0.06/day
- This is the single biggest cost in Design B

### Config
```yaml
hybridMemory:
  retrieval:
    reranking:
      enabled: false           # off by default, Design B feature
      model: 'sonnet'          # mid-tier for quality
      candidateCount: 50       # how many to re-rank
      outputCount: 20          # how many to keep
```

## Priority
P1 within Design B — Highest cost component but significant quality improvement.

markus-lassfolk:
## 📋 Gap Audit (PR-aware) — Issue #161

### ✅ Implemented
- LLM re-ranking service with configurable model (PR #223, merged)
- Config: `reranking.enabled`, `model`, `candidateCount`, `outputCount` (PR #223)
- Disabled by default as per spec (PR #223)
- Integration with retrieval pipeline post-RRF fusion (PR #223)

### ⚠️ Gaps / Needs Verification
- Re-ranking actually invoked in live retrieval flow — needs wiring confirmation
- No v2 PR exists — first-pass may have left runtime integration incomplete
- Prompt template for LLM re-ranking — verify matches spec

### 📊 Completion: ~70% (needs wiring verification)

_PR-aware audit by Maeve — 2026-03-08_

PR #258, opened 2026-03-08
feat: Milestone A+B complete feature implementation (#144-#206)

### LINKED ISSUE #144: Feature: Future-date decay protection for reminders and deadlines
## Summary

Facts containing future dates (reminders, deadlines, upcoming events) should have their decay halted until that date passes. Currently, a reminder set 3 weeks out could decay below retrieval threshold before it fires.

## Motivation

If I store "dentist appointment on March 20" on March 1, and the decay class is `active`, confidence will erode over 19 days. By the time March 20 arrives, the fact might be too faint to surface in retrieval — defeating the purpose.

Total Recall handles this by detecting future dates in observations and freezing their decay until the date passes. Simple, effective.

## Proposed behaviour

1. On `memory_store`, detect if the text contains a future date (regex + lightweight NLP)
2. If found, set a `decayFreezeUntil` timestamp on the fact
3. `memory_prune` skips confidence decay for facts where `NOW < decayFreezeUntil`
4. After the date passes, normal decay resumes

## Edge cases

- Multiple dates in one fact → use the latest one
- Recurring events → don't freeze (they'll be re-stored)
- Vague dates ("next week", "soon") → best-effort parsing, skip if ambiguous

## Source

Idea from [Total Recall](https://github.com/gavdalf/total-recall)'s future-date protection logic.

markus-lassfolk:
## Summary

Full feature implementation for Milestone A and B issues. All gaps identified via Scholar audit and fixed by Forge.

### Milestone A
- **#144** — Future-date decay protection (recurring event detection)
- **#150** — Episodic Event Log (passive observer wiring, 90-day archival)
- **#153** — ONNX local embedding provider (migration tooling)

### Milestone B  
- **#158** — Multi-model embedding registry (RRF merge)
- **#159** — Contextual variants (contextual-search + contextual-means)
- **#160** — Query expansion (conditional mode, threshold, rule-based aliases)
- **#161** — LLM re-ranking (wired into retrieval pipeline)
- **#162** — Verification store (auto-classify, JSON backup, continuous reverification)
- **#163** — Provenance (recursive chain building, memory_provenance tool)
- **#206** — Document ingestion (hash dedup, LLM vision, folder ingestion)

### Stats
- 12 squash-merged commits
- ~1,500+ lines of new code
- All tests passing at each merge point

Closes #144, #150, #153, #158, #159, #160, #161, #162, #163, #206

---

> [!NOTE]
> **Medium Risk**
> Touches the OpenAI proxy and SQLite write path to instrument every chat completion with token/cost logging, plus adds new CLI/reporting and query APIs; failures are best-effort but regressions could affect LLM call performance or DB stability.
> 
> **Overview**
> **Adds opt-out LLM cost tracking (Issue #270).** Introduces `CostTracker` tables (`llm_cost_log`, `llm_savings_log`), a static model pricing table, and instruments the multi-provider OpenAI proxy to record per-call tokens/costs (with feature labeling via AsyncLocalStorage or prompt heuristics); adds `hybrid-mem cost-report` plus nightly pruning and verify output showing recent spend and mode-based cost estimates.
> 
> **Expands dashboard-oriented data access and feature wiring.** `FactsDB` gains decay-class stats, a paginated/filterable `listForDashboard` with optional FTS search, and link edge export for graph views; frustration detection now captures assistant turns and reuses a cached tool-effectiveness store, while cross-agent learning and tool-effectiveness are refactored to use public config/types and raw DB access safely.
> 
> **Config/docs updates.** Extends reinforcement config options, documents new `implicitFeedback`, `closedLoop`, and `errorReporting` settings/privacy details, adds `costTracking` config parsing/types, and tweaks frustration default `decayRate` (0.9→0.85).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit eabff0ea212468757a25ad38b529f8bf3a8f6024. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


extensions/memory-hybrid/tools/memory-tools.ts
```diff
@@ -403,6 +494,13 @@ export function registerMemoryTools(
             asOfSec ?? undefined,
             cfg.aliases?.enabled ? aliasDb : null,
             cfg.clusters,
+            embeddingRegistry ?? null,
```
markus-lassfolk: ### 🔴 Undefined embeddingRegistry in memory tools

**Severity**

`embeddingRegistry` is referenced here, but it is not part of `PluginContext` or destructured from `ctx`. This will fail type-checking or throw a `ReferenceError` at runtime. Pass `embeddingRegistry` into `registerMemoryTools` (and add it to `PluginContext`) or remove its usage.

extensions/memory-hybrid/tools/memory-tools.ts
```diff
@@ -833,6 +956,30 @@ export function registerMemoryTools(
           };
         }
 
+        const explicitVerificationTier = (verificationTier ?? "").trim().toLowerCase();
+
+        const maybeAutoVerify = (factId: string, factText: string, autoTags: string[], autoEntity?: string | null, autoKey?: string | null, autoValue?: string | null) => {
+          if (!cfg.verification.enabled || !verificationStore) return;
```
markus-lassfolk: ### 🔴 Undefined verificationStore in memory_store flow

**Severity**

`verificationStore` is used here but is not in `PluginContext` and is never pulled from `ctx`. This will crash at runtime when the code path executes. Add `verificationStore` to the tool context and wire it from `registerTools`, or gate the logic to avoid referencing an undefined variable.

extensions/memory-hybrid/services/retrieval-orchestrator.ts
```diff
@@ -429,304 +429,341 @@ export async function runRetrievalPipeline(
...
+    queryExpander && typeof (queryExpander as QueryExpander).getMode === "function"
+      ? queryExpander.getMode()
+      : (queryExpander ? "always" : "off");
+
+  if (expanderMode === "conditional") {
+    const alias =
+      queryExpander && typeof (queryExpander as QueryExpander).getRuleBasedAlias === "function"
+        ? queryExpander.getRuleBasedAlias(query)
+        : null;
+    const initial = await runOnce({ useLlm: false, variants: alias ? [alias] : [], skipReranking: true });
```
markus-lassfolk: ### 🟡 Reranking skipped in conditional expansion

**Severity**

In conditional query-expansion mode you call `runOnce(..., skipReranking: true)` and return early when the initial score is above the threshold. That means LLM reranking is silently disabled even when `rerankingConfig.enabled` is true. If the intent is to avoid LLM use entirely, this should be explicit in config/docs; otherwise consider reranking the initial results before returning.

extensions/memory-hybrid/backends/event-log.ts
```diff
@@ -192,6 +214,117 @@ export class EventLog {
...
+                severity: "info",
+                subsystem: "event-log",
+              });
+            }
+          }
+        }
+
+        const lineStream = Readable.from((function* (self: EventLog) {
+          if (existingLines) {
+            yield existingLines;
```
markus-lassfolk: ### 🟡 Archive idempotency risk and memory spike

**Severity**

`existingLines` loads the entire existing archive into memory and re-writes it along with rows selected from the DB. If a previous run wrote the archive but crashed before deletion, this will append duplicate rows. Consider streaming append-only with dedupe (by `id`) or writing a new archive from DB rows only after a successful delete/transaction to keep the operation idempotent.

extensions/memory-hybrid/services/embeddings.ts
```diff
@@ -38,6 +44,466 @@ const GOOGLE_EMBEDDING_BASE_URL = "https://generativelanguage.googleapis.com/v1b
...
+function splitRepoAndRevision(input: string): { repo: string; revision: string } {
+  const at = input.lastIndexOf("@");
+  if (at > 0 && at < input.length - 1) {
+    return { repo: input.slice(0, at), revision: input.slice(at + 1) };
+  }
+  return { repo: input, revision: "main" };
+}
+
+async function downloadFile(url: string, destPath: string): Promise<void> {
+  const resp = await fetch(url);
```
markus-lassfolk: ### 🟡 Unbounded model downloads

**Severity**

`downloadFile` fetches ONNX models without timeouts, size limits, or checksum validation. A bad or very large model can hang startup or fill disk. Consider adding a fetch timeout, max byte cap, and optional checksum/ETag validation for known models.

markus-lassfolk (commented):
## 🤖 Codex Code Review — PR #258

Summary of findings:
- 2 critical runtime errors in memory tools: undefined embeddingRegistry and verificationStore
- Reranking is skipped in conditional expansion mode even when enabled
- Event log archiving is not idempotent and can duplicate entries
- ONNX model download lacks timeout/size/integrity safeguards

Signed: 🤖 Codex Reviewer

extensions/memory-hybrid/services/contextual-variants.ts
```diff
@@ -169,13 +201,19 @@ export class VariantGenerationQueue {
...
+          for (const variantType of VARIANT_TYPES) {
+            try {
+              const variants = await this.generator.generateVariants(
+                item.text,
+                item.category,
+                variantType,
+              );
+              if (variants.length > 0) {
+                await this.onVariantsGenerated(item.factId, variantType, variants);
+              }
```
markus-lassfolk: **Issue #159 - helper classes exist, but I could not find the runtime integration**

`ContextualVariantGenerator` / `VariantGenerationQueue` are implemented here, but I could not find any normal runtime path that instantiates the queue or calls `enqueue(...)` after storing a fact. In the live `memory_store` flow I only found canonical embedding storage plus aliases; I did not find automatic contextual variant generation/storage being triggered for write-time facts.

That means the issue is still only partially fulfilled: the service and tests exist, but the behavior described by #159 ('generate contextual variants at fact storage time and store them alongside canonical embeddings') does not appear to run in the actual write path on `develop`.

extensions/memory-hybrid/tools/memory-tools.ts
```diff
@@ -403,6 +494,13 @@ export function registerMemoryTools(
             asOfSec ?? undefined,
             cfg.aliases?.enabled ? aliasDb : null,
             cfg.clusters,
+            embeddingRegistry ?? null,
+            factsDb,
+            null, // queryExpander (not wired here)
+            null, // embedFn (not wired here)
```
markus-lassfolk: **Issue #160 - query expansion is still disabled in the live recall path**

`runRetrievalPipeline()` supports query expansion, but the normal `memory_recall` path here still passes `null` for both `queryExpander` and `embedFn`, and I could not find any runtime `new QueryExpander(...)` outside tests. So even though the service/orchestrator code exists, the behavior requested by #160 never actually activates during normal recall.

In practice this means the issue is still incomplete: setting query-expansion config does not make `memory_recall` expand and fan out variants on the real runtime path.

extensions/memory-hybrid/setup/register-tools.ts
```diff
@@ -186,7 +201,12 @@ export function registerTools(ctx: ToolsContext, api: ClawdbotPluginApi): void {
...
   // Document ingestion tool (opt-in, requires Python + markitdown)
   if (cfg.documents.enabled && pythonBridge) {
-    registerDocumentTools({ factsDb, vectorDb, cfg, embeddings, pythonBridge }, api);
+    registerDocumentTools({ factsDb, vectorDb, cfg, embeddings, pythonBridge, openai }, api);
+  }
+
+  // Verification tools (Issue #162)
+  if (cfg.verification.enabled && verificationStore) {
+    registerVerificationTools({ factsDb, verificationStore }, api);
   }
```
markus-lassfolk: **Issue #162 - explicit management tools are wired, but retrieval-time integrity checking still appears to be missing**

This review wiring adds the verification tools, which is useful. But #162 also required verified/critical facts to be integrity-checked on retrieval so corruption is detected during normal reads. I could not find any call from `memory_recall`, auto-recall, or other normal retrieval paths into `verificationStore.getVerified()` / `checkIntegrity()` before returning facts.

So the store and tools exist, but one of the core behavioral guarantees from #162 ('checksum-verified on retrieval') still does not appear to happen automatically in the live read path.

extensions/memory-hybrid/services/passive-observer.ts
```diff
@@ -571,7 +576,46 @@ export async function runPassiveObserver(
...
+          extractionMethod: 'passive',
+          extractionConfidence: fact.importance,
         })
+        if (opts.provenanceService) {
+          try {
+            opts.provenanceService.addEdge(stored.id, {
+              edgeType: "DERIVED_FROM",
+              sourceType: "event_log",
+              sourceId: sessionId,
+              sourceText: fact.text,
```
markus-lassfolk: **Issue #163 - provenance is still not fully trustworthy / comprehensive**

The provenance edge recorded here uses `sourceType: "event_log"`, but the `sourceId` being stored is `sessionId`, not the actual event-log row id that gets appended later. That breaks the issue's fact-to-source tracing goal because `memory_provenance` cannot reliably resolve this back to a specific episodic event via `eventLog.getById(...)`.

I also checked document ingestion and found that `memory_ingest_document` stores chunks without provenance edges/columns, so the issue's 'every fact' requirement is still not satisfied across all write paths.

extensions/memory-hybrid/tools/document-tools.ts
```diff
@@ -27,17 +28,632 @@ export interface DocumentToolsContext {
   vectorDb: VectorDB;
   cfg: HybridMemoryConfig;
   embeddings: EmbeddingProvider;
+  openai: OpenAI;
   pythonBridge: PythonBridge;
 }
 
+type ProgressTracker = {
+  steps: string[];
```
markus-lassfolk: **Issue #206 - substantial implementation exists, but the checklist is still not fully met**

The folder/document ingestion work here is substantial, but the implementation still falls short of two explicit checklist items from #206. First, the code accumulates progress messages and returns them at the end, but that is not the same as progress callbacks / streaming progress during a long-running ingest. From the runtime surface I reviewed, callers only receive the final response after completion.

Second, the bridge robustness is still partial: startup health-check/restart exists, but `PythonBridge.convert()` does not retry a failed in-flight conversion after worker exit/timeout. So the issue's 'health check + restart on crash' requirement is only partially met right now.

markus-lassfolk:
Addressed review feedback in commit 09a0881:

**Cursor Bugbot:** Backup fields (action, id, ts, nextVerification, previousVersionId) in verification-store; shared `expandTilde` in `utils/path.ts`; per–variant-type rate limit in contextual-variants.

**Wiring:** `embeddingRegistry` and `verificationStore` passed into `registerMemoryTools` and destructured in memory-tools; `detectFutureDate` import and `cfg.futureDateProtection ?? { enabled: false }` guard.

**Conditional reranking:** When above threshold, initial results are now reranked when `rerankingConfig.enabled` (retrieval-orchestrator).

**Event log archive:** Stream existing file (readline + gunzip), dedupe by id, idempotent merge; no full-file buffer.

**ONNX download:** 120s timeout, 500MB max, `ReadableStream` type, size guard while streaming.

**Embedding migration:** `getCount`/`getBatch` when available (batched); fallback to `getAll` for mocks; meta updated when `autoMigrate=false` but config changed.

**Provenance:** Passive observer uses event id from `eventLog.append()` as `sourceId` (eventLog already in opts and passed from plugin-service).

**verified_facts:** FK `fact_id REFERENCES facts(id) ON DELETE CASCADE` + migration for existing DBs.

**Query expansion:** `QueryExpander` and `embedFn` wired in `memory_recall` when `queryExpansion.enabled`.

**Tests:** vision test uses bridge that throws for `conversion_failed`; listDue test makes new version due; embedding-migration uses mock without getCount/getBatch.

Remaining open items from the thread (e.g. #150 layer order, #153 plugin.json schema, #159 runtime enqueue, #162 retrieval-time integrity, #163 document provenance, #206 progress callbacks) are left for follow-up.

extensions/memory-hybrid/services/passive-observer.ts
```diff
@@ -571,7 +576,46 @@ export async function runPassiveObserver(
...
+              factId: stored.id,
+            })
+          }
+        }
+
+        // Write episodic event (Issue #150): record what was learned to Layer 1
+        if (opts.eventLog) {
+          try {
+            opts.eventLog.append({
+              sessionId,
```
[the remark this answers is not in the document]
markus-lassfolk: Fixed in commit 932b250: reversed write order so event_log.append() runs before factsDb.store(); uses the returned event ID as provenance sourceId.

extensions/memory-hybrid/setup/register-tools.ts
markus-lassfolk: Fixed in commit 41449ef: memory_recall now calls verificationStore.checkIntegrity(factId) for each retrieved fact when verification is enabled; tampered facts are flagged with [⚠️ TAMPERED] prefix in the output.

extensions/memory-hybrid/services/embedding-migration.ts
```diff
@@ -0,0 +1,297 @@
...
+            text: fact.text,
+            vector: vec,
+            importance: fact.importance ?? 0.5,
+            category: fact.category,
+          });
+          factsDb.setEmbeddingModel(fact.id, embeddings.modelName);
+          migrated++;
+        } else {
+          skipped++;
+        }
```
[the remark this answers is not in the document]
markus-lassfolk: Fixed in commit 41449ef: hasDuplicate check now runs BEFORE vectorDb.delete(); the old vector is only removed when isDuplicate is false, so we never drop a valid entry and then discover it would have been a duplicate.

PR #391, opened 2026-03-12
fix: Bug: Ollama model OOM — qwen3:8b requires 18.2 GiB but only ~8 GiB available (#387)

### LINKED ISSUE #387: Bug: Ollama model OOM — qwen3:8b requires 18.2 GiB but only ~8 GiB available
## Source
GlitchTip issues: HYBRID-MEMORY-2C, 2B, 2A, 29 (7 events total, 2026-03-12)

## Problem
Ollama local model (`qwen3:8b`) used for auto-recall HyDE generation and classifier-batch operations requires ~18.2 GiB system memory but only ~8 GiB is available on the host. This causes a `500` error from the Ollama API.

## Affected Operations
- `auto-recall-hyde-generation` (subsystem: auto-recall) — 3 issues
- `classify-batch` (subsystem: classifier) — 1 issue

## Stack
```
extensions/openclaw-hybrid-memory/node_modules/openai/src/client.ts:738 in OpenAI.makeRequest
```

## Expected Behavior
When the local Ollama model doesn't have enough memory, the plugin should:
1. Detect the OOM condition and fall back to a cloud provider (Gemini Flash, etc.)
2. Or use a smaller local model that fits in available RAM
3. At minimum, log a clear warning rather than sending errors to GlitchTip repeatedly

## Environment
- Release: openclaw-hybrid-memory@2026.3.110
- Host: ~8 GiB available RAM
- Model: qwen3:8b (requires 18.2 GiB)

markus-lassfolk:
## Summary

- Adds `isOllamaOOM()` detection helper to identify Ollama out-of-memory responses (e.g. `model requires more system memory (18.2 GiB) than is available (8.0 GiB)`)
- OOM errors are now suppressed from GlitchTip reporting across all affected subsystems (`auto-recall-hyde-generation`, `classify-batch`, `chatComplete`, `withLLMRetry`, `chatCompleteWithRetry`)
- OOM short-circuits retry logic in `withLLMRetry` (no delay, falls through to next fallback model immediately)
- Ollama embedding circuit breaker now trips immediately on OOM HTTP 500 response (instead of waiting for 3 connection failures)
- Clear user-facing warning logged when Ollama OOM is detected during HyDE generation

## Test plan

- [x] `isOllamaOOM()` matches all known Ollama OOM error message patterns
- [x] `withLLMRetry` does not retry on OOM (called exactly once, no delay)
- [x] OOM not reported to GlitchTip in `withLLMRetry`
- [x] `chatCompleteWithRetry` falls through to fallback model immediately on OOM
- [x] OOM not reported to GlitchTip when all models fail with OOM
- [x] Ollama embedding circuit breaker trips immediately on OOM HTTP 500 body
- [x] Generic HTTP 500 (non-OOM) does NOT trip circuit breaker immediately
- [x] All 3228 existing tests pass, TypeScript clean

Closes #387

---

> [!NOTE]
> **Medium Risk**
> Changes error classification and retry/circuit-breaker behavior around LLM and embedding calls; mistakes could suppress real errors or alter fallback behavior under failure conditions.
> 
> **Overview**
> Improves handling of Ollama out-of-memory failures by adding `isOllamaOOM()` and treating these as *expected* non-bug conditions across chat, HyDE query expansion, and auto-classification.
> 
> OOM (and other transient 5xx/404/timeout) errors are now suppressed from GlitchTip reporting, OOM short-circuits `withLLMRetry` retries to immediately fall through to fallback models, and `chatCompleteWithRetry` surfaces a user warning when all models fail due to OOM.
> 
> For embeddings, Ollama HTTP 500 responses that indicate OOM now trip the circuit breaker immediately (disabling the endpoint for the cooldown period), with new tests covering OOM detection, retry suppression, fallback behavior, and circuit-breaker changes.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 339dad15290333d09c7c5360e2b04c3e0d3389c4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/extensions/memory-hybrid/services/chat.ts
+++ b/extensions/memory-hybrid/services/chat.ts
@@ -512,9 +512,15 @@ export async function chatCompleteWithRetry(opts: {
         phase: "fallback-exhausted",
       });
     }
-  } else if (finalIs500 || finalIsOOM) {
+  } else if (finalIsOOM) {
+    // #387: OOM is a persistent condition (model too large for RAM), not transient — warn user to use smaller model
+    pendingWarnings?.add(
+      `⚠️ Memory plugin: LLM model requires more memory than available (OOM). ` +
+      `Consider using a smaller model or configuring a cloud fallback. ` +
+      `Run: openclaw hybrid-mem verify --test-llm`
+    );
+  } else if (finalIs500) {
     // #302: 500 server errors are transient — don't report to GlitchTip; request will be retried naturally
-    // #387: OOM is expected when model too large for available RAM — not a bug
   } else if (finalIs404) {
     // #303: model not found across all fallbacks = misconfigured model name — surface to user, skip Sentry
     pendingWarnings?.add(


extensions/memory-hybrid/services/chat.ts
```diff
@@ -105,6 +108,58 @@ function is500Like(err: unknown): boolean {
...
+ * Unified transient LLM error detection helper.
+ * Checks if an error is transient (OOM, 5xx, 404, timeout) and should not be reported to GlitchTip.
+ * Centralizes the four-condition check used across hooks.ts and auto-classifier.ts.
+ *
+ * Exported so other modules (lifecycle/hooks, auto-classifier) can suppress
+ * capturePluginError for transient LLM failures without duplicating the logic.
+ */
+export function isTransientLLMError(err: unknown): boolean {
+  if (!(err instanceof Error)) return false;
+  return (
```
markus-lassfolk: ### 🟡 `isTransientLLMError` not used in `chatComplete` — divergence risk

**Warning**

`chatComplete()` (around line 245-257) still uses an inline `isTransient` check with overlapping but non-identical logic: it checks abort, timeout, ECONNREFUSED, 500, and now OOM individually. Meanwhile `isTransientLLMError()` centralizes OOM + 500 + 404 + timeout. These two checks cover different subsets and will diverge silently if future error categories are added to only one.

Consider either:
1. Refactoring `chatComplete` to use `isTransientLLMError()` (plus its additional abort/ECONNREFUSED checks), or
2. Adding a code comment cross-referencing the two locations so maintainers update both.

Not urgent — the current logic is correct — but this is a latent maintenance trap.

extensions/memory-hybrid/services/chat.ts
```diff
@@ -105,6 +108,58 @@ function is500Like(err: unknown): boolean {
...
+ * Extracted helper so both Error objects and raw HTTP body strings can be checked.
+ * Ollama returns HTTP 500 with a body like:
+ *   "model requires more system memory (18.2 GiB) than is available (8.0 GiB)"
+ *
+ * Exported so embeddings.ts can check raw HTTP body strings for OOM patterns.
+ */
+export function isOllamaOOMMessage(msg: string): boolean {
+  const lower = msg.toLowerCase();
+  return (
+    lower.includes("model requires more system memory") ||
```
markus-lassfolk: ### 🟡 `isOllamaOOMMessage` regex has non-trivial false-positive surface

**Warning**

The pattern `/requires\s+[\d.]+\s*gib/i` matches any error message mentioning GiB requirements — not only Ollama. A database allocation error like `operation requires 4.0 GiB of disk` or a CUDA/GPU error like `model requires 8.0 GiB VRAM` would match.

Since `isOllamaOOM` → `isTransientLLMError` → suppress GlitchTip, a false positive here means silently swallowing a non-Ollama error report.

Suggestion: tighten with an Ollama-context anchor, e.g.:
```ts
/requires\s+[\d.]+\s*gib.*(?:than is available|system memory)/i.test(msg)
```
Or require `lower.includes('ollama')` as a precondition (though Ollama errors don't always self-identify). Low risk today since these helpers are only called in Ollama-adjacent code paths, but worth hardening as usage spreads.

extensions/memory-hybrid/services/embeddings.ts
```diff
@@ -828,7 +828,19 @@ export class OllamaEmbeddingProvider implements EmbeddingProvider {
       }
       if (!resp.ok) {
         const body = await resp.text().catch(() => "");
-        throw new Error(`Ollama embed failed: HTTP ${resp.status} ${resp.statusText}${body ? ` — ${body.slice(0, 200)}` : ""}`);
+        const errMsg = `Ollama embed failed: HTTP ${resp.status} ${resp.statusText}${body ? ` — ${body.slice(0, 200)}` : ""}`;
+        // OOM: trip circuit breaker immediately — retrying the same model won't free memory.
+        const isOOM = isOllamaOOMMessage(body);
+        if (isOOM) {
+          circuit.disabledUntil = Date.now() + OLLAMA_COOLDOWN_MS;
+          circuit.failCount = OLLAMA_MAX_FAILS;
```
markus-lassfolk: ### 🟢 Circuit breaker direct state mutation couples to internals

**Suggestion**

The OOM handler directly sets `circuit.disabledUntil` and `circuit.failCount = OLLAMA_MAX_FAILS` to trip the breaker. This works but couples to the circuit-breaker implementation details. If the circuit-breaker logic changes (e.g., adds a state enum, exponential backoff, or logging), this manual trip won't benefit.

Consider adding a `circuit.trip(reason?: string)` method that encapsulates this — reusable if other immediate-trip scenarios arise. Not blocking, just a maintainability improvement.

extensions/memory-hybrid/services/chat.ts
```diff
@@ -92,8 +92,11 @@ export function is404Like(err: unknown): boolean {
  * Unified 5xx / internal server error detection helper.
```
markus-lassfolk: ### 🔴 Early return prevents detection of HTTP status objects

**Critical**

In `isTransientLLMError`, returning `false` early via `if (!(err instanceof Error))` fails to pass non-Error objects (e.g., `{ status: 500 }`) to `is500Like(err)` and `is404Like(err)`, which are specifically designed to inspect those properties. This causes transient server errors to be incorrectly reported to GlitchTip.

**Fix:** Remove the early return and allow `err` to flow into the helper functions. Only require an `Error` instance when checking the `.message` property.

`isTransientLLMError` — reference to the specific code.

markus-lassfolk: ### 🟡 `isOllamaOOM` ignores string and object errors

**Warning**

In `isOllamaOOM`, if an embedding provider or raw `fetch` call throws an OOM error as a raw string or plain object (e.g., `{ message: "model requires..." }`), the `if (!(err instanceof Error))` check will instantly fail to detect it. *(Note: The test `"returns false for non-Error values"` enforces this incorrect behavior and should be updated too)*.

**Fix:** Extract the message from the `unknown` error safely (e.g. `typeof err === 'object' && 'message' in err`) instead of returning `false` instantly.

`isOllamaOOM` — reference to the specific code.

extensions/memory-hybrid/lifecycle/hooks.ts
```diff
@@ -24,7 +24,7 @@ import type { EventLog } from "../backends/event-log.js";
 import type { AliasDB } from "../services/retrieval-aliases.js";
```
markus-lassfolk: ### 🔴 Eagerly wrapping `err` strips HTTP status properties

**Critical**

In the `catch (err)` block for HyDE generation, `err` is wrapped into an `Error` instance via `new Error(String(err))` *before* calling `isTransientLLMError`. If `err` is a raw HTTP response object like `{ status: 500 }`, it converts to `new Error("[object Object]")`. This strips the `status` property that `is500Like` relies on, causing the error to be incorrectly reported to GlitchTip.

**Fix:** Pass the original `err` to `isTransientLLMError(err)` so its internal helpers can inspect the raw object properties, then use `hydeErr` only for `capturePluginError`.

extensions/memory-hybrid/services/auto-classifier.ts
```diff
@@ -12,6 +12,7 @@ import type { FactsDB } from "../backends/facts-db.js";
 import { getMemoryCategories, setMemoryCategories, isValidCategory } from "../config.js";
```
markus-lassfolk: ### 🔴 Eagerly wrapping `err` strips HTTP status properties

**Critical**

Similar to `hooks.ts`, wrapping `err` into `discErr` and `classifyErr` via `new Error(String(err))` strips object properties like `status`. If `err` is an object like `{ status: 500 }`, `isTransientLLMError(discErr)` receives an Error with message `"[object Object]"` and no `status` property, causing it to return `false`.

**Fix:** Pass the original `err` to `isTransientLLMError(err)` in both the `discoverCategoriesFromOther` and `classifyBatch` functions.

extensions/memory-hybrid/services/embeddings.ts
```diff
@@ -828,7 +828,23 @@ export class OllamaEmbeddingProvider implements EmbeddingProvider {
...
+          /requires\s+[\d.]+\s*gib/i.test(body) ||
+          /\boom:/i.test(body);
+        if (isOOM) {
+          circuit.disabledUntil = Date.now() + OLLAMA_COOLDOWN_MS;
+          circuit.failCount = OLLAMA_MAX_FAILS;
+          console.warn(
+            `memory-hybrid: Ollama model OOM (${this.modelName}) — model requires more memory than available. ` +
+            `Circuit breaker tripped; disabling endpoint ${this.endpoint} for 5min. ` +
+            `Consider using a smaller model or configuring a cloud embedding fallback.`
+          );
```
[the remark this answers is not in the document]
markus-lassfolk: Fixed in 8c6fe13. Added OLLAMA_COOLDOWN_MIN = Math.round(OLLAMA_COOLDOWN_MS / 60_000) immediately after the OLLAMA_COOLDOWN_MS constant, then replaced both hardcoded '5min' strings in the circuit-breaker log messages with the computed value. The log messages now stay accurate if the constant changes.

extensions/memory-hybrid/services/auto-classifier.ts
markus-lassfolk: Fixed in 8c6fe13. Both catch blocks in auto-classifier.ts (category-discovery-batch and classify-batch) now pass the original err to isTransientLLMError before coercion. The coerced Error (discErr / classifyErr) is kept for capturePluginError which needs an Error instance. A comment was added explaining the pattern.

PR #467, opened 2026-03-13
fix(python-bridge): add requirements.txt and startup dep check for markitdown

### LINKED ISSUE #422: GlitchTip: Error: Python RPC error -32000: markitdown is not installed. Run: pip install markitdown
## GlitchTip Error Report

**Source:** HYBRID-MEMORY-30
**Level:** error
**Occurrences:** 1
**First seen:** 2026-03-12
**Last seen:** 2026-03-12
**GlitchTip link:** http://192.168.1.99:8000/hybrid-memory/issues/337
**GlitchTip ID:** 337

### Error
```
Error: Python RPC error -32000: markitdown is not installed. Run: pip install markitdown
```

### Auto-filed
This issue was automatically created by the GlitchTip → GitHub sync pipeline.
Review the GlitchTip event details for stack traces and context.

markus-lassfolk:
Closes #422

## Summary

Fixes the Python RPC error when `markitdown` is not installed.

### Changes

1. **`requirements.txt`** — New file at `extensions/memory-hybrid/scripts/requirements.txt` listing `markitdown` as a required Python dependency. Users can now run `pip install -r extensions/memory-hybrid/scripts/requirements.txt` during setup.

2. **`PythonBridge.checkDependencies()`** — New method using `spawnSync` to verify required Python packages are installed. Returns `{ ok, missing }` without throwing.

3. **Startup warning** — `index.ts` now calls `checkDependencies()` immediately after creating the `PythonBridge` instance (when `documents.enabled`). Missing packages are logged as a warning with the exact `pip install` command, so the issue is visible in logs at startup rather than only on the first document conversion attempt.

---

> [!NOTE]
> **Medium Risk**
> Adds a synchronous Python import check during plugin registration when `documents.enabled`, which could affect startup behavior if the Python binary is slow/misconfigured, but changes are localized and non-fatal (warnings only).
> 
> **Overview**
> When `documents.enabled`, the plugin now **proactively verifies Python prerequisites at startup** by calling `PythonBridge.checkDependencies()` and logging clear warnings for either a missing/failed Python executable or missing packages, including an explicit `pip install` command.
> 
> `PythonBridge` adds a new synchronous `checkDependencies()` method (using `spawnSync`) to detect missing imports (currently `markitdown`), and the PR introduces `extensions/memory-hybrid/scripts/requirements.txt` to document/install the required Python dependency. Tests extend `python-bridge.test.ts` to cover the new dependency-check outcomes.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2c200c59009c342699a01d8acc9fd064d792a181. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/extensions/memory-hybrid/services/python-bridge.ts
+++ b/extensions/memory-hybrid/services/python-bridge.ts
@@ -240,7 +240,7 @@ export class PythonBridge {
         encoding: "utf8",
       });
       if (result.error) {
-        return { ok: false, missing: [], spawnError: result.error };
+        return { ok: false, missing, spawnError: result.error };
       }
       if (result.status !== 0) {
         const output = (result.stderr ?? "") + (result.stdout ?? "");
@@ -252,7 +252,7 @@ export class PythonBridge {
           // surface a more actionable message.
           return {
             ok: false,
-            missing: [],
+            missing,
             spawnError: new Error(
               `Python import check failed (status=${result.status}): ${output.slice(0, 200)}`,
             ),


extensions/memory-hybrid/services/python-bridge.ts
```diff
@@ -222,4 +222,27 @@ export class PythonBridge {
...
+   * Runs synchronously (fast — just a python import check) so it can be called
+   * at plugin startup to surface missing dependencies early rather than on first use.
+   *
+   * @returns Object with `ok` flag and list of any `missing` packages.
+   */
+  checkDependencies(): { ok: boolean; missing: string[] } {
+    const required = ["markitdown"];
+    const missing: string[] = [];
+    for (const pkg of required) {
+      const result = spawnSync(this.pythonPath, ["-c", `import ${pkg}`], {
```
markus-lassfolk: ### 🟡 spawnSync failure ≠ import failure — misleading missing-package report

**Warning**

`result.status !== 0` is true for **any** non-zero exit: Python binary not found, permission denied, or a genuine `ImportError`. If `this.pythonPath` is wrong or the binary is missing, every package in `required` will be falsely flagged as missing.

Consider checking `result.error` first:

```ts
if (result.error) {
  // spawn failed entirely — log separately, don't add to missing[]
  return { ok: false, missing: [] };
}
if (result.status !== 0) {
  missing.push(pkg);
}
```

This prevents the misleading `Run: pip install markitdown` message when the real problem is a misconfigured `pythonPath`.

extensions/memory-hybrid/index.ts
```diff
@@ -391,6 +391,19 @@ const memoryHybridPlugin = {
     // Initialized lazily — PythonBridge only spawns the subprocess on first convert() call
     pythonBridge = cfg.documents.enabled ? new PythonBridge(cfg.documents.pythonPath) : null;
 
+    // Eagerly check Python dependencies at startup so missing packages surface
+    // immediately (in logs) rather than on first document conversion (issue #422).
+    if (pythonBridge) {
+      const { ok, missing } = pythonBridge.checkDependencies();
+      if (!ok) {
+        const pkgs = missing.join(", ");
+        api.logger.warn(
```
markus-lassfolk: ### 🟡 pip install hint may target wrong Python environment

**Warning**

The warning message says `Run: pip install markitdown` but the dependency check runs against `this.pythonPath`. If the user runs bare `pip`, it may install into a different environment and the check will still fail.

Fix: use the same Python binary:

```ts
`Run: ${this.pythonPath} -m pip install ${missing.join(" ")}  (see extensions/memory-hybrid/scripts/requirements.txt)`
```

This guarantees the package ends up where the bridge will look for it.

extensions/memory-hybrid/services/python-bridge.ts
```diff
@@ -222,4 +222,27 @@ export class PythonBridge {
...
+        timeout: 5_000,
+        encoding: "utf8",
+      });
+      if (result.status !== 0) {
+        missing.push(pkg);
+      }
+    }
+    return { ok: missing.length === 0, missing };
+  }
 }
```
markus-lassfolk: ### 🟢 No unit test for checkDependencies()

**Suggestion**

The new method has no test coverage. A minimal test (mock `spawnSync` returning status 0 vs 1 vs error) would prevent regressions and document the expected behaviour for the two edge cases (spawn failure vs import failure).

markus-lassfolk (commented):
## 🦊 Pipeline Council Review — PR #467

**fix(python-bridge): add requirements.txt and startup dep check for markitdown**

The change is well-scoped and solves the right problem: surfacing missing markitdown early instead of on first document conversion. Core logic is sound. Two warnings worth addressing before merge, one suggestion.

Findings: 2 warnings, 1 suggestion.

Signed: 🦊 Maeve Pipeline Council

extensions/memory-hybrid/index.ts
markus-lassfolk: Fixed in 0e83ae8: changed hint to `python3 -m pip install markitdown` to use the same Python interpreter used for the check.

PR #526, opened 2026-03-16
fix: [Ralph Test] Passive observer dedup uses O(n*m) brute-force cosine similarity in (#499)

### LINKED ISSUE #499: [Ralph Test] Passive observer dedup uses O(n*m) brute-force cosine similarity instead of vector index
## Goal Addressed
**Engineering Goal 2: Cost & Latency Efficiency** — aggressively optimize LLM vs local heuristics.
**Product Goal 4: Autonomous Maintenance** — clean, consolidate, and reflect in the background.

## Files/Modules
- `extensions/memory-hybrid/services/passive-observer.ts` — dedup loop at ~L410-430

## Current Flaw
The passive observer extracts facts from session transcripts on a timer. For each extracted fact, it computes an embedding and then runs a **linear scan** against all recent fact vectors (up to 50) using `dotProductSimilarity`. With `f` extracted facts per run (typically 5-15 per chunk, across multiple chunks), and `m` recent vectors, the total is `f * m` dot-product operations per run.

More critically, this dedup check **does not use the existing LanceDB vector index at all**. LanceDB already stores all fact vectors and supports approximate nearest-neighbor search in sub-millisecond time. The passive observer instead maintains its own parallel `recentVectors[]` array, pre-computes embeddings via `embedBatch()`, and does manual similarity — duplicating what the vector DB already does natively.

Additionally, the `recentFacts` pool is hard-capped at 50 entries from the last 7 days. A busy agent storing 100+ facts/day means the dedup window covers less than half a day, allowing duplicates from older sessions to slip through.

## Proposed Fix
1. **Use `vectorDb.search(vec, 1, threshold)` for dedup**: Replace the brute-force loop with a single LanceDB query per extracted fact. LanceDB's IVF index makes this O(log n) instead of O(m). This eliminates the need for `embedBatch()` on the entire recent facts pool (saving 50 embedding API calls per run).
2. **Remove `recentVectors[]` / `recentFactIds[]` arrays**: The vector DB is the single source of truth. Maintaining a parallel in-memory copy violates separation of concerns.
3. **For the reinforcement path**: When a near-duplicate is found via LanceDB, look up the matched fact ID and call `boostConfidence()` as before. LanceDB search already returns the fact ID.
4. **Expand the dedup window naturally**: LanceDB searches the entire corpus, not just the last 50 facts, so the 7-day/50-fact blind spot disappears.
5. **Keep `dotProductSimilarity` as a fallback**: If LanceDB is unavailable (circuit breaker open), fall back to the current brute-force approach with a smaller pool.

markus-lassfolk:
## Summary

- Removes Phase 2 brute-force dedup: `getRecentFacts()` + `embedBatch()` on up to 50 recent facts + in-memory `dotProductSimilarity` linear scan — **O(f×m) per run**
- Replaces with a single `vectorDb.search(vec, 1, similarityThreshold)` call per extracted fact — **O(log n)** via LanceDB's ANN index
- Eliminates ~50 unnecessary embedding API calls per passive-observer run (Engineering Goal 2: Cost & Latency Efficiency)
- Removes the parallel `recentVectors[]`/`recentFactIds[]` in-memory arrays — LanceDB is the single source of truth (Engineering Goal 3: Strict Separation of Concerns)
- **Bugfix**: Normalizes vectors before deduplication search and storage to ensure the cosine-to-L2 conversion formula is valid for all embedding providers (Ollama, Google, etc.)
- Uses normalized vectors for consistent ANN search and L2 distance metric
- Reinforcement path preserved: matched fact ID from `dupeResults[0].entry.id` → `boostConfidence()`
- Graceful fallback on search failure: treats as non-duplicate so a few extra duplicates are acceptable rather than crashing
- **Fixed test assertion**: Updated test to expect the converted L2-based threshold value instead of raw cosine similarity threshold
- **Bugfix**: Restored intra-batch deduplication for dry-run mode by maintaining `dryRunVectors[]` array to track facts that would be stored in the current run, ensuring dry-run accurately previews real-run behavior
- **Bugfix**: Replaced inline dot product with `dotProductSimilarity()` utility from `reflection.ts` to add missing length guard and reduce code duplication
- **Bugfix**: Restored confidence reinforcement capability in intra-batch dedup path by tracking fact IDs alongside vectors in `recentVectors[]` array

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## Quality Checklist

- [x] `cd extensions/memory-hybrid && npx tsc --noEmit` passes with no errors
- [x] `cd extensions/memory-hybrid && npm run lint` passes with no warnings
- [x] `cd extensions/memory-hybrid && npm test` passes (all tests green)
- [x] No secrets, credentials, or API keys committed
- [x] No `console.log` debug statements left in production code

## Testing

- [x] Unit tests added / updated
- [x] Rewrote dedup test as end-to-end (mocked `vectorDb.search` returning a hit)
- [x] Added 5 new tests in `runPassiveObserver — LanceDB dedup (Issue #499)` describe block:
  - `vectorDb.search` called once per extracted fact
  - `deduplicationThreshold` forwarded as `minScore` (test now correctly expects converted L2-based value)
  - Reinforcement path boosts matched fact via `boostConfidence`
  - Search failure → graceful fallback (fact still stored)
  - `getRecentFacts` is never called (Phase 2 eliminated)
- [x] Updated `confidence-reinforcement.test.ts` mock to include `search`
- [x] All 74 passive-observer tests pass

### Bug Fix: Vector Normalization

The L2-to-cosine conversion formula `1 / (1 + sqrt(2*(1-cosine)))` is only valid for unit-length (normalized) vectors. While OpenAI embeddings are pre-normalized, Ollama and Google providers may return unnormalized vectors. Without normalization, the L2 distance is scaled by vector magnitude, making the converted threshold miscalibrated and causing duplicates to slip through undetected.

The fix normalizes vectors before:
1. Deduplication search via `vectorDb.search()`
2. Storage in LanceDB via `vectorDb.store()`
3. Dry-run intra-batch comparison

This ensures the conversion formula is mathematically valid for all embedding providers and maintains consistent L2 distance metrics in LanceDB.

### Bug Fix: Inline Dot Product Duplication

The intra-batch dedup code manually reimplemented the dot product loop instead of using the existing `dotProductSimilarity()` utility from `reflection.ts`. The utility includes a defensive `a.length !== b.length` guard that returns 0 on dimension mismatch, which the inline version lacked. While dimension mismatches are unlikely within the same embedding model, using the canonical implementation reduces maintenance burden and ensures consistent behavior.

### Bug Fix: Confidence Reinforcement in Intra-Batch Dedup

The intra-batch dedup fallback path (used when `vectorDb.store()` fails) was unable to perform confidence reinforcement because fact IDs were not tracked alongside vectors. The fix restores this capability by changing `recentVectors` from `number[][]` to `Array<{ vector: number[]; factId: string }>`, enabling `boostConfidence()` calls when duplicates are detected in the same batch.

Closes #499

<div><a href="https://cursor.com/agents/bc-e13d083f-1146-41af-a260-038946a98e70"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-e13d083f-1146-41af-a260-038946a98e70"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/extensions/memory-hybrid/tests/passive-observer.test.ts
+++ b/extensions/memory-hybrid/tests/passive-observer.test.ts
@@ -567,10 +567,16 @@ describe("runPassiveObserver — LanceDB dedup (Issue #499)", () => {
     const vectorDb = makeVectorDb(); // search returns [] — no duplicates
     const cfg = makeConfig({ sessionsDir });
 
+    // Use distinct vectors so intra-batch dedup doesn't catch the second fact
+    const embeddings = makeEmbeddings();
+    embeddings.embed
+      .mockResolvedValueOnce([0.1, 0.2, 0.3])
+      .mockResolvedValueOnce([0.9, 0.1, 0.0]);
+
     const result = await runPassiveObserver(
       makeFactsDb() as never,
       vectorDb as never,
-      makeEmbeddings() as never,
+      embeddings as never,
       {} as never,
       cfg,
       ["fact"],


extensions/memory-hybrid/services/passive-observer.ts
```diff
@@ -27,7 +27,6 @@ import { chunkTextByChars } from "../utils/text.js";
 import { loadPrompt, fillPrompt } from "../utils/prompt-loader.js";
```
markus-lassfolk: ### 🟢 Excellent algorithmic optimization

**Suggestion**

Replacing the O(n*m) brute-force similarity scan with an O(log n) LanceDB ANN index query is a great scaling improvement and neatly fulfills the PR's engineering goals by leveraging the vector DB directly instead of pulling embeddings into memory.

PR #567, opened 2026-03-17
fix: improvement: recall pipeline vector-step timeout does not cancel HyDE LLM call — (#558)

### LINKED ISSUE #558: improvement: recall pipeline vector-step timeout does not cancel HyDE LLM call — model stays busy after timeout
## Problem
In `services/recall-pipeline.ts`, the vector search step has a 30s timeout implemented via `Promise.race`. When the timeout fires, `directiveAbort.abort()` is called. However, the HyDE generation (`chatCompleteWithRetry`) is called before the `directiveAbort.signal` is checked, and the signal is only passed to `chatCompleteWithRetry` — it is not guaranteed to actually cancel the in-flight HTTP request.

Concretely:
1. HyDE call starts at t=0 (chatCompleteWithRetry with `signal: directiveAbort.signal`)
2. Vector timeout fires at t=30s — `directiveAbort.abort()` called
3. But `chatCompleteWithRetry` may already be in a retry loop: `withLLMRetry` re-issues the HTTP call after a backoff without re-checking the abort signal between retries
4. The aborted `vectorStepPromise` still awaits in the background — the finally block re-attaches `.catch()` to suppress noise, but the LLM call and embedding call continue executing in the background

The result: a timed-out recall step wastes LLM tokens on a HyDE call whose result will be discarded, and may block the embedding provider during the next recall invocation.

## Proposed Solution
1. Check `directiveAbort.signal.aborted` at the start of the embed step (after HyDE) and throw early if aborted.
2. In `withLLMRetry`, re-check the abort signal before each retry iteration (not just on fetch — add a check before the `await` in the retry loop).
3. Add a test: simulate a slow HyDE response that exceeds the timeout, assert that the embedding provider is NOT called after abort.

## Files Affected
- `extensions/memory-hybrid/services/recall-pipeline.ts`
- `extensions/memory-hybrid/services/chat.ts` (`withLLMRetry` retry loop)
- `extensions/memory-hybrid/tests/recall-pipeline.test.ts`

## Acceptance Criteria
- [ ] `embeddings.embed` is not called when `directiveAbort.signal` is already aborted
- [ ] `withLLMRetry` checks `signal.aborted` before each retry attempt
- [ ] Existing recall timeout tests still pass
- [ ] Test: abort signal fired before embed step — embed not called (mock assertion)

## Effort Estimate
M

markus-lassfolk:
## Summary

- Add abort guard before `embeddings.embed` in `recall-pipeline.ts`: checks `directiveAbort.signal.aborted` immediately after HyDE completes and throws early if the 30s vector-step timeout has already fired, preventing a wasted embedding provider call whose result will be discarded
- Add pre-backoff signal check in `withLLMRetry` (`chat.ts`): re-checks `opts.signal.aborted` before each retry backoff sleep so a parent-level abort (e.g. vector-step timeout) unblocks the retry loop immediately
- Add 2 new tests in `recall-pipeline.test.ts`: one using fake timers to fast-forward past the 30s timeout and assert `embed` is not called, plus a regression test verifying `embed` IS still called when HyDE fails for a non-abort reason

## Test plan
- [ ] All existing recall-pipeline tests pass (16/16)
- [ ] New test: `does not call embeddings.embed when vector-step timeout fires while HyDE is running` — passes
- [ ] New test: `still calls embeddings.embed with raw query when HyDE fails non-abort` — passes
- [ ] No TypeScript errors in changed files
- [ ] No console.log added

Closes #558

---

> [!NOTE]
> **Medium Risk**
> Touches recall-timeout and LLM retry/backoff behavior; mistakes could change cancellation semantics or retry timing, impacting recall latency and provider call volume.
> 
> **Overview**
> Prevents wasted embedding calls in the recall pipeline by **bailing out before `embeddings.embed`** when the vector-step timeout abort has already fired after a long-running HyDE request.
> 
> Makes `withLLMRetry` backoff **abort-aware** (and adds 429-specific backoff honoring `Retry-After`), so parent aborts/timeouts can interrupt retry sleeps immediately instead of waiting out the delay.
> 
> Adds regression tests covering the timeout-abort case (ensuring `embed` is not called) and a non-abort HyDE failure case (ensuring raw-query embedding still runs).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e3cc21b112214ce7f50db8ffd3ec07c26ff62cc5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/extensions/memory-hybrid/services/chat.ts
+++ b/extensions/memory-hybrid/services/chat.ts
@@ -526,19 +526,24 @@ export async function withLLMRetry<T>(
       } else {
         delay = Math.pow(3, attempt) * 1000; // 1s, 3s, 9s
       }
-      await new Promise((resolve, reject) => {
-        const timeout = setTimeout(resolve, delay);
+      // Abort-aware backoff sleep: if the signal fires while we are waiting, reject immediately
+      // instead of sleeping through the full delay. The listener is removed on normal resolve to
+      // prevent leaks; the { once: true } option is not relied on alone for cleanup.
+      await new Promise<void>((resolve, reject) => {
+        const onAbort = () => {
+          clearTimeout(timeout);
+          const reason = opts!.signal!.reason;
+          const msg = reason instanceof Error ? reason.message : reason != null ? String(reason) : "Aborted";
+          const abortError = new Error(msg);
+          abortError.name = "AbortError";
+          reject(abortError);
+        };
+        const timeout = setTimeout(() => {
+          opts?.signal?.removeEventListener("abort", onAbort);
+          resolve();
+        }, delay);
         if (opts?.signal) {
-          const onAbort = () => {
-            clearTimeout(timeout);
-            const reason = opts.signal!.reason;
-            const abortError =
-              reason instanceof Error ? reason : new Error(reason != null ? String(reason) : "Aborted");
-            abortError.name = "AbortError";
-            reject(abortError);
-          };
           if (opts.signal.aborted) {
-            clearTimeout(timeout);
             onAbort();
           } else {
             opts.signal.addEventListener("abort", onAbort, { once: true });
@@ -591,7 +596,8 @@ export async function chatCompleteWithRetry(opts: {
   for (let i = 0; i < modelsToTry.length; i++) {
     if (signal?.aborted) {
       const reason = (signal as AbortSignal).reason;
-      const abortError = reason instanceof Error ? reason : new Error(reason != null ? String(reason) : "Aborted");
+      const msg = reason instanceof Error ? reason.message : reason != null ? String(reason) : "Aborted";
+      const abortError = new Error(msg);
       abortError.name = "AbortError";
       throw abortError;
     }
--- a/extensions/memory-hybrid/services/recall-pipeline.ts
+++ b/extensions/memory-hybrid/services/recall-pipeline.ts
@@ -165,7 +165,9 @@ export async function runRecallPipelineQuery(
         // The HyDE call above may have completed just before the abort — we must not
         // waste an embedding provider call whose result will be discarded.
         if (directiveAbort.signal.aborted) {
-          throw new Error(`recall pipeline timed out after ${VECTOR_STEP_TIMEOUT_MS}ms`);
+          const abortError = new Error(`recall pipeline timed out after ${VECTOR_STEP_TIMEOUT_MS}ms`);
+          abortError.name = "AbortError";
+          throw abortError;
         }
 
         const vector =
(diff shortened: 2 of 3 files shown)


extensions/memory-hybrid/services/chat.ts
```diff
@@ -525,7 +526,25 @@ export async function withLLMRetry<T>(
...
+              reason instanceof Error ? reason : new Error(reason != null ? String(reason) : "Aborted");
+            abortError.name = "AbortError";
+            reject(abortError);
+          };
+          if (opts.signal.aborted) {
+            clearTimeout(timeout);
+            onAbort();
+          } else {
+            opts.signal.addEventListener("abort", onAbort, { once: true });
+          }
```
markus-lassfolk: ### 🟢 Abort listener not cleaned up on normal resolve

**Minor**

When the backoff timer fires normally, the `onAbort` listener stays registered on the signal until abort fires (harmless due to `{ once: true }`) or the signal is GC'd. For cleanliness, remove the listener in the resolve path:

```ts
const timeout = setTimeout(() => {
  opts.signal!.removeEventListener("abort", onAbort);
  resolve();
}, delay);
```

extensions/memory-hybrid/services/chat.ts
```diff
@@ -525,7 +526,25 @@ export async function withLLMRetry<T>(
...
+          const onAbort = () => {
+            clearTimeout(timeout);
+            const reason = opts.signal!.reason;
+            const abortError =
+              reason instanceof Error ? reason : new Error(reason != null ? String(reason) : "Aborted");
+            abortError.name = "AbortError";
+            reject(abortError);
+          };
+          if (opts.signal.aborted) {
+            clearTimeout(timeout);
```
markus-lassfolk: ### 🔴 Event Listener Leak

**Bug**

The `onAbort` handler is added to `opts.signal` with `addEventListener("abort", onAbort, { once: true })`, but if the `setTimeout` fires first (i.e., the delay completes normally before any abort), the listener is **never removed**. The listener will linger on the AbortSignal for the lifetime of the controller, which can cause memory leaks and potentially a spurious `reject()` on a settled Promise if an abort fires later in the same retry loop.

**Fix:** Clean up the event listener when the timeout resolves normally:

```typescript
await new Promise<void>((resolve, reject) => {
  const onAbort = () => {
    clearTimeout(timeout);
    const reason = opts.signal!.reason;
    const abortError =
      reason instanceof Error ? reason : new Error(reason != null ? String(reason) : "Aborted");
    abortError.name = "AbortError";
    reject(abortError);
  };

  const timeout = setTimeout(() => {
    opts.signal?.removeEventListener("abort", onAbort); // clean up
    resolve();
  }, delay);

  if (opts?.signal) {
    if (opts.signal.aborted) {
      clearTimeout(timeout);
      onAbort();
    } else {
      opts.signal.addEventListener("abort", onAbort, { once: true });
    }
  }
});
```

PR #575, opened 2026-03-17
fix: improvement: capture-utils.ts and auto-capture.ts duplicate SENSITIVE_PATTERNS — (#555)

### LINKED ISSUE #555: improvement: capture-utils.ts and auto-capture.ts duplicate SENSITIVE_PATTERNS — risk of divergence
## Problem
There are two separate `SENSITIVE_PATTERNS` arrays defined in the codebase that serve overlapping purposes:

1. **`services/auto-capture.ts`** — used for credential detection and vault routing (8 patterns including AWS key regex, private key header, connection strings with passwords)
2. **`services/capture-utils.ts`** — used for `shouldCapture` to filter out sensitive text from memory capture (7 patterns, simpler — no AWS key regex, no connection string pattern)

These two arrays have diverged silently. `capture-utils.ts`'s `SENSITIVE_PATTERNS` is missing the `AKIA[0-9A-Z]{16}` AWS key pattern, the `-----BEGIN .*PRIVATE KEY` private key pattern, and the `://user:pass@host` connection string pattern that `auto-capture.ts` catches. This means an AWS key in a message would NOT be blocked from memory capture by `shouldCapture`, even though the vault logic would catch it separately.

Additionally, `consolidation.ts` imports from `auto-capture.ts` (the more complete set), suggesting the intent was a single source of truth.

## Proposed Solution
1. Make `auto-capture.ts` the single source of truth for `SENSITIVE_PATTERNS` (it's already the more complete version).
2. Re-export or import `SENSITIVE_PATTERNS` in `capture-utils.ts` from `auto-capture.ts`.
3. Add a test asserting both AWS key and private key patterns are blocked by `shouldCapture`.

## Files Affected
- `extensions/memory-hybrid/services/capture-utils.ts`
- `extensions/memory-hybrid/services/auto-capture.ts`
- `extensions/memory-hybrid/tests/credentials-auto-capture.test.ts` (add cases)

## Acceptance Criteria
- [ ] Single `SENSITIVE_PATTERNS` definition (in `auto-capture.ts`); `capture-utils.ts` imports from it
- [ ] `shouldCapture` blocks AWS key (`AKIA...`), private key header, and connection strings with passwords
- [ ] Tests cover the three patterns that were missing from `capture-utils.ts`

## Effort Estimate
S

markus-lassfolk:
## Summary

- Makes `auto-capture.ts` the single source of truth for `SENSITIVE_PATTERNS` by adding the three patterns that were unique to `capture-utils.ts` (`/\bbearer\b/i`, `/\bauthorization\b/i`, `/\bcredentials?\b/i`)
- Removes the duplicate local `SENSITIVE_PATTERNS` array from `capture-utils.ts` and imports from `auto-capture.ts` instead
- Adds three new tests in `credentials-auto-capture.test.ts` verifying that `shouldCapture` now correctly blocks AWS keys (`AKIA...`), private key headers (`-----BEGIN ... PRIVATE KEY`), and connection strings with embedded passwords

## Test plan

- [x] All 3429 existing tests pass
- [x] `npx tsc --noEmit` is clean
- [x] New tests confirm the three previously-missing patterns are now blocked by `shouldCapture`

Closes #555

---

> [!NOTE]
> **Low Risk**
> Low risk: refactors duplicate sensitive-pattern lists into a shared export and tightens capture filtering; main risk is unintended over-blocking/under-blocking of auto-capture based on regex behavior.
> 
> **Overview**
> Makes `auto-capture.ts` the **single source of truth** for sensitive-text detection by expanding `SENSITIVE_PATTERNS` (now includes `bearer`, `authorization`, and `credentials`) and reusing it from `capture-utils.ts` instead of maintaining a second list.
> 
> Introduces `CAPTURE_FILTER_PATTERNS` to keep consolidation’s more-specific `token is` check while ensuring `shouldCapture` and `isCredentialLike` still block broader `token` formats (e.g. `token=` / `token:`). Adds tests to verify `shouldCapture` rejects AWS keys, private key headers, connection strings with embedded passwords, and common token assignment formats even when a memory trigger matches.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9605ad854ae40213b87062015d5a3c85c1130064. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/extensions/memory-hybrid/services/auto-capture.ts
+++ b/extensions/memory-hybrid/services/auto-capture.ts
@@ -30,9 +30,7 @@ export const SENSITIVE_PATTERNS = [
 ];
 
 /** Patterns for capture filtering - uses broader /token/i instead of /token\s+is/i for security */
-export const CAPTURE_FILTER_PATTERNS = SENSITIVE_PATTERNS.map((pattern) =>
-  pattern.source === /token\s+is/i.source && pattern.flags === /token\s+is/i.flags ? /token/i : pattern,
-);
+export const CAPTURE_FILTER_PATTERNS = [...SENSITIVE_PATTERNS.slice(0, 3), /token/i, ...SENSITIVE_PATTERNS.slice(4)];
 
 /** Patterns that suggest a credential value - for auto-detect prompt to store */
 const CREDENTIAL_PATTERNS: Array<{ regex: RegExp; type: string; hint: string }> = [


markus-lassfolk (commented):
**Review of PR #575**

This PR cleanly resolves the duplicated  issue (#555) by centralizing them in .

**Key observations:**
1.  **Security & Correctness:** The introduction of  is a smart approach. Using the broader  for auto-capture filtering () ensures we don't accidentally leak tokens formatted as  or  into general memory. Keeping  in the base  prevents over-aggressive filtering during memory consolidation.
2.  **Completeness:** Adding , , and  to the shared list ensures consistent redaction/filtering across the board.
3.  **Test Coverage:** The newly added tests in  accurately verify the behavior of  against the consolidated patterns, specifically confirming that AWS keys, private keys, and varied token formats are correctly blocked.
4.  **False Positives:** The broader  pattern in  might block legitimate (non-secret) text containing the word "token" from being auto-captured. However, in the context of memory capture, prioritizing security over a missed auto-capture is the right tradeoff.

The implementation is solid and improves the security posture of the auto-capture mechanism.

🔮 Gemini Council Review — PR #575

markus-lassfolk (commented):
Nice cleanup overall — making `auto-capture.ts` the single source of truth removes the drift that caused #555, and the new tests cover the previously-missed high-risk cases (AWS keys, private keys, embedded-password URLs, `token=` and `token:` forms).

Main concern: broadening the capture filter from `/token\s+is/i` to `/token/i` is a meaningful behavioral change, not just a security fix. It will now suppress any auto-capture candidate containing the substring `token`, including benign notes like token budgets/limits, OAuth token rotation reminders, tokenizer/debugging notes, etc. That tradeoff may be acceptable, but right now the tests only prove the new positive-blocking cases; they do not lock in the intended false-positive boundary.

Two concrete suggestions:
1. Add at least one negative regression test showing whether benign `token` mentions are intentionally blocked or should still capture.
2. Consider making `CAPTURE_FILTER_PATTERNS` explicit instead of splice-by-index (`slice(0, 3)` / `slice(4)`), because that coupling silently breaks if `SENSITIVE_PATTERNS` is reordered later.

So: security coverage is better, but I would like the broader `/token/i` behavior to be documented/tested more explicitly before calling this fully settled.

🤖 GPT Council Review — PR #575

PR #579, opened 2026-03-17
fix: improvement: auto-classifier LLM category discovery fires on every cron invocati (#557)

### LINKED ISSUE #557: improvement: auto-classifier LLM category discovery fires on every cron invocation with no per-run cooldown
## Problem
`services/auto-classifier.ts` → `discoverCategoriesFromOther` is called whenever `autoClassify` cron runs AND `config.suggestCategories === true`. The guard condition is:

```ts
if (others.length < MIN_OTHER_FOR_DISCOVERY) return [];  // MIN = 15
```

There is no time-based cooldown. If you have ≥15 `"other"` facts, every cron invocation (potentially every 6–24h) will call the LLM once per 25-fact batch. With 100 `"other"` facts, that's 4 LLM calls per run. If categories are discovered but none of the facts are reclassified (e.g. confidence threshold not met), the count stays ≥15 and the LLM fires again next run.

The `discoveredCategoriesPath` file is read to avoid re-adding already-known categories, but the LLM prompt fires regardless — it has no awareness of when it last ran.

## Proposed Solution
Add a timestamp entry to `discoveredCategoriesPath` (or a separate sidecar `.last-discovery.json`) recording the last successful discovery run. Before calling the LLM, check: if last run was within `discoveryIntervalHours` (config, default 72h), skip. This prevents O(facts/batchSize) LLM calls on every cron tick when categories are already settled.

Also add a config key `autoClassify.discoveryIntervalHours` (default: 72) with documentation.

## Files Affected
- `extensions/memory-hybrid/services/auto-classifier.ts`
- `extensions/memory-hybrid/config/types/features.ts` (add `discoveryIntervalHours` to AutoClassifyConfig)
- `extensions/memory-hybrid/config/parsers/` (parse new config key)

## Acceptance Criteria
- [ ] `discoverCategoriesFromOther` checks last-run timestamp and skips if within cooldown
- [ ] Default cooldown is 72h; configurable via `autoClassify.discoveryIntervalHours`
- [ ] Last-run timestamp written after successful discovery; not written on error
- [ ] Test: two calls within cooldown window → second call skips LLM entirely (mock asserts called once)

## Effort Estimate
M

markus-lassfolk:
## Summary

- Add `discoveryIntervalHours` config key to `AutoClassifyConfig` (default: 72h) to control cooldown between category discovery LLM runs
- Implement cooldown check in `discoverCategoriesFromOther`: reads a `.last-run.json` sidecar file and skips LLM calls entirely if elapsed time is within the interval
- Write last-run timestamp to sidecar after all LLM batches complete (even when 0 new categories found); not written on error so failures are retried
- `discoveryIntervalHours = 0` disables cooldown for testing/manual runs
- Export `getLastDiscoveryPath()` for deterministic sidecar path derivation and testing
- Add 13 unit tests in `auto-classifier-cooldown.test.ts` covering: cooldown enforcement, sidecar write/read, timestamp not overwritten on skip, expired cooldown re-fires LLM, and zero-interval override

## Test plan

- [x] All 3454 existing tests pass (`npm test`)
- [x] TypeScript clean (`npx tsc --noEmit`)
- [x] 13 new tests specifically for cooldown behavior all pass
- [x] Test: two calls within cooldown window → second call skips LLM (mock asserts called once) ✅

Closes #557

---

> [!NOTE]
> **Medium Risk**
> Changes auto-classifier cron behavior by persisting and enforcing a discovery cooldown via a sidecar timestamp file, which could affect when/if new categories are suggested and introduces new filesystem I/O paths.
> 
> **Overview**
> Prevents the auto-classifier’s *category discovery* step from calling the LLM on every cron tick by adding a configurable cooldown (`autoClassify.discoveryIntervalHours`, default `72`, `0` disables).
> 
> Discovery now reads/writes a `.last-run.json` sidecar timestamp (derived via exported `getLastDiscoveryPath`) and skips discovery when within the interval; the timestamp is only written when at least one discovery batch succeeds so failures are retried. Adds Vitest coverage for cooldown enforcement, sidecar behavior, and config parsing.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f2619f73655712e71662c41f988b4fc073b523b9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/extensions/memory-hybrid/services/auto-classifier.ts
+++ b/extensions/memory-hybrid/services/auto-classifier.ts
@@ -118,6 +118,7 @@ async function discoverCategoriesFromOther(
 
   const existingCategories = new Set(getMemoryCategories());
   const labelToIds = new Map<string, string[]>();
+  let anyBatchSucceeded = false;
 
   for (let i = 0; i < others.length; i += DISCOVERY_BATCH_SIZE) {
     const batch = others.slice(i, i + DISCOVERY_BATCH_SIZE);
@@ -140,6 +141,7 @@ async function discoverCategoriesFromOther(
       const jsonMatch = content.match(/\[[\s\S]*\]/);
       if (!jsonMatch) continue;
       const labels: unknown[] = JSON.parse(jsonMatch[0]);
+      anyBatchSucceeded = true;
       for (let j = 0; j < Math.min(labels.length, batch.length); j++) {
         const raw = typeof labels[j] === "string" ? (labels[j] as string) : "";
         const label = normalizeSuggestedLabel(raw);
@@ -174,9 +176,10 @@ async function discoverCategoriesFromOther(
     for (const id of ids) factsDb.updateCategory(id, label);
   }
 
-  // Write last-run timestamp after a successful discovery run (even if no new categories were created).
+  // Write last-run timestamp only if at least one batch succeeded (even if no new categories were created).
   // This prevents the LLM from firing again on the next cron tick when categories are settled.
-  if (intervalHours > 0) {
+  // If all batches failed, don't write the timestamp so discovery is retried on the next run.
+  if (intervalHours > 0 && anyBatchSucceeded) {
     const lastRunPath = getLastDiscoveryPath(discoveredCategoriesPath);
     await writeLastDiscoveryTimestamp(lastRunPath, Date.now());
   }


markus-lassfolk (commented):
## 🔮 Gemini Council Review — PR #579

**Overall: Solid implementation with good test coverage. A few minor issues worth addressing.**

---

### ✅ Strengths

- Cooldown logic is well-encapsulated and the `getLastDiscoveryPath` export keeps path derivation DRY and testable.
- `anyBatchSucceeded` flag is the right approach — only persisting the timestamp on actual LLM success prevents permanent suppression after transient failures.
- 13 + WAL helper tests provide solid regression coverage.
- Sidecar file errors are silently swallowed (`return null`), which is the safe degradation path.

---

### 🐛 Issues / Concerns

#### 1.  leaks into the production bundle

`wal-helpers.ts` exports this symbol unconditionally. Since there's no tree-shaking guarantee in this package, the test-only reset hook ships to production consumers. Options:
- Guard with `if (process.env.NODE_ENV === 'test')` before export, or
- Move it to a dedicated `wal-helpers.testing.ts` re-export that tests import directly.

#### 2. Type inconsistency: required vs. optional 

`AutoClassifyConfig` in `config/types/retrieval.ts` declares `discoveryIntervalHours: number` (required), but all three inline config shapes in `auto-classifier.ts` (`discoverCategoriesFromOther`, `runClassifyForCli`, `runAutoClassify`) declare it as `discoveryIntervalHours?: number` (optional). The `??` fallback handles it correctly at runtime, but the type contract is misleading — callers of the internal functions see it as optional when the public config type says it's always present.

Consider either making it `discoveryIntervalHours?: number` in the public type (with the default living only in the parser), or aligning the inline types to also be required.

#### 3. No README / user-facing docs update

The new `autoClassify.discoveryIntervalHours` config key isn't documented in the README or any config reference. Users upgrading won't know this knob exists. A brief mention in the configuration docs (default 72h, 0 to disable) would be helpful.

#### 4. Minor: timestamp written via two separate  calls

The cooldown check reads `Date.now()` at line ~107, and `writeLastDiscoveryTimestamp` reads it again later (~line 183). In practice the drift is negligible for 72h granularity, but pinning `const nowMs = Date.now()` once at the top of the function would be slightly cleaner and easier to mock in tests.

#### 5. No file locking on sidecar — concurrent cron runs can both fire discovery

If two cron ticks overlap (e.g. a slow previous run still running when the next fires), both processes will read the sidecar as absent/expired and both will hit the LLM. This is an inherent limitation of file-based state without atomic locking. Not a blocker for the stated goal (reduce LLM calls per cron tick), but worth a `// NOTE: no file lock — concurrent runs may both fire; acceptable for this use case` comment so future readers don't try to "fix" it with complex locking.

---

### 📝 Nits

- The constant `DEFAULT_DISCOVERY_INTERVAL_HOURS = 72` in `auto-classifier.ts` and the default in `parseAutoClassifyConfig` are both hardcoded to 72 independently. A single source of truth (e.g. import the constant from the service into the parser) would prevent them drifting.

---

**Summary:** Issues #1 (test hook in production bundle) and #3 (missing docs) are the most actionable before merge. The rest are low-risk polish items. Great overall direction — this is exactly the kind of cost-control guard that prevents LLM runaway in long-running agents.

PR #611, opened 2026-03-18
fix: WAL remove() does a full O(n) parse after every single removal (#588)

### LINKED ISSUE #588: WAL remove() does a full O(n) parse after every single removal
## Problem

In `extensions/memory-hybrid/backends/wal.ts`, the `remove()` method appends a remove marker and then **immediately calls `readAll()`** to check if the WAL is empty:

```typescript
remove(id: string): void {
  const line = JSON.stringify({ op: "remove", id }) + "\n";
  appendFileSync(this.walPath, line, "utf-8");
  this.fsyncAfterWrite();
  if (this.readAll().length === 0) this.clear(); // <-- full file parse every time
}
```

`readAll()` parses the entire WAL file line-by-line, building a Set of removed IDs and filtering entries. For a WAL with N entries, each `remove()` is O(N). If you remove M entries in sequence (e.g. during WAL replay), the total cost is O(M × N).

## Suggested fix

Option A (simple): Track entry count in memory:
```typescript
private entryCount = 0;

write(entry: WALEntry): void {
  // ... existing logic ...
  this.entryCount++;
}

remove(id: string): void {
  // ... append remove marker ...
  this.entryCount--;
  if (this.entryCount <= 0) this.clear();
}
```

Option B (safer): Only auto-clear during `pruneStale()`, not on every `remove()`. Let the periodic cleanup handle compaction.

## Files affected

- `extensions/memory-hybrid/backends/wal.ts`

markus-lassfolk:
## Summary

- Adds `activeIds: Set<string>` to `WriteAheadLog`, seeded once in the constructor via a single `readAll()` parse
- `write()` adds the entry ID to the set; `remove()` deletes it and checks `size === 0` to decide on compaction — **no file I/O or parsing on the hot path**
- `clear()` and `pruneStale()` keep the set in sync after compaction
- Constructor's initial `readAll()` is wrapped in try-catch so a bad path (e.g. a directory used in error-handling tests) still lets construction succeed

**Before:** every `remove()` appended a tombstone then called `readAll()` — a full two-pass file parse — to check if the log was empty. Batch deletions were O(n²) in aggregate.

**After:** the compaction check is O(1) (Set.size). Batch deletions are O(n).

Closes #588

## Test plan

- [x] Existing WAL tests all pass (33/33)
- [x] New: `"does not call readAll() during remove — O(1) compaction check"` — spies on `readAll` and asserts it is never called from `remove()`
- [x] New: `"batch removals do not trigger any readAll() call"` — 10-entry batch delete, zero `readAll()` calls
- [x] New: `"auto-clears WAL after all entries removed via batch removes"` — file is deleted after last entry removed
- [x] New: `"removes non-existent ID without clearing when other entries exist"` — non-existent ID removal is a no-op on the set, WAL file untouched
- [x] `npx tsc --noEmit` — clean

---

> [!NOTE]
> **Medium Risk**
> Touches persistence logic for the write-ahead log; while the change is localized, any divergence between the new in-memory index and on-disk WAL could cause incorrect clearing/retention of entries.
> 
> **Overview**
> Eliminates the per-`remove()` full WAL re-parse by introducing an in-memory `activeIds` set in `WriteAheadLog`, seeded once on construction and maintained on `write()`, `remove()`, `clear()`, and after `pruneStale()` compaction.
> 
> `remove()` now performs an O(1) emptiness check via `activeIds.size` to decide when to `clear()` the WAL file, and the test suite adds coverage to ensure `readAll()` is never called during single or batch removals and that clearing behavior remains correct.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 18c837a04b6375cc62ab975d6474506858433c7e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


markus-lassfolk (commented):
## 🔮 Gemini Council Review — PR #611

**fix: WAL remove() O(n²) → O(n) via in-memory activeIds Set**

### Summary
This is a clean, well-targeted performance fix. The core insight is correct: replacing a per-`remove()` `readAll()` parse with an in-memory `Set<string>` eliminates the quadratic cost of batch deletions. The implementation is consistent — `activeIds` is properly maintained across all four mutation paths (`write`, `remove`, `clear`, `pruneStale`).

### Strengths ✅
- **O(1) hot path**: `activeIds.size === 0` check is a genuine improvement; batch deletions now O(n) total instead of O(n²)
- **Constructor safety**: The `try-catch` seeding is the right call — bad paths (e.g. a directory) shouldn't prevent construction
- **`pruneStale()` sync**: Reassigning `activeIds = new Set(valid.map(e => e.id))` after compaction is correct and keeps the set fresh
- **Test coverage**: Four focused tests covering the key behavioral guarantees, including the non-existent ID edge case
- **No regression risk on error paths**: `activeIds` mutations happen after successful `appendFileSync`, so error-path state remains consistent

### Observations / Minor Concerns ⚠️

**1. Spurious tombstone writes for non-existent IDs (minor inefficiency)**
`remove()` always appends a tombstone to disk even when the ID is not in `activeIds`. This was the original behavior too, but now that we have the set, an early guard would be cheap and prevent WAL file bloat:
```ts
// Optional optimization: skip disk write if ID was never active
if (!this.activeIds.has(id)) return;
```
Not a correctness issue, but worth considering given the PR's performance focus.

**2. Duplicate-write invariant (pre-existing, not introduced here)**
If `write()` is called twice with the same ID, the set correctly deduplicates but both entries land on disk. `readAll()` would return duplicates until the next compaction. This is pre-existing behavior — noting it here for awareness, not as a blocker.

**3. `readAll` spy coupling in tests (low risk)**
The new tests spy on `readAll` as a public method. This is appropriate and pragmatic for this optimization, but it does create implementation-coupled tests — if `readAll` is ever legitimately called during these operations for other reasons, the tests would break. Consider whether a comment noting this coupling would help future maintainers.

**4. No test for constructor seeding with existing WAL file**
The constructor seeds `activeIds` from an existing WAL file on disk, but there's no explicit test for a `WriteAheadLog` instantiated over a pre-populated WAL (i.e., the crash-recovery / restart scenario). The existing tests likely cover this implicitly, but an explicit test for "new WAL instance over pre-existing WAL file has correct activeIds" would close the loop on the seeding correctness claim.

### Risk Assessment
**Low-Medium.** The in-memory set introduces a new failure mode — desync between set and disk — but the implementation correctly guards against this on all mutation paths. The constructor try-catch means a seeding failure silently initializes an empty set, which could cause a WAL with existing entries to never auto-clear after all removals are done (the set thinks it's empty from construction, but IDs still exist on disk). This is handled by the file still being parseable via `readAll()` and the next `pruneStale()` resync.

### Verdict
**Approve with minor suggestions.** The performance fix is correct and well-tested. The spurious-tombstone optimization and the constructor-seeding test are nice-to-haves, not blockers. Ready to merge.

markus-lassfolk (commented):
## 🤖 GPT Council Review — PR #611

**fix: WAL remove() O(n²) → O(1) via in-memory activeIds Set**

This is a clean, well-motivated performance fix. The before/after complexity story is compelling and the implementation is straightforward. Here's my take:

### ✅ Strengths

- **Correct seeding**: One-time `readAll()` in the constructor properly initializes the Set from disk state, so restarts are safe.
- **All mutation paths covered**: `write()`, `remove()`, `clear()`, and `pruneStale()` all keep `activeIds` in sync. The `pruneStale()` reassignment (`new Set(valid.map(...))`) is especially correct — it mirrors exactly what gets written back to disk after compaction.
- **Good defensive constructor**: The try-catch around the initial `readAll()` handles bad paths gracefully without silently swallowing errors on the hot path.
- **Test quality**: Spying on `readAll` to assert zero-call contracts is exactly the right approach. All four new tests are meaningful.

### ⚠️ Issues Worth Noting

**1. Tombstone writes for non-existent IDs (minor optimization miss)**

In `remove()`, the tombstone is unconditionally appended to disk before checking the Set:
```ts
appendFileSync(this.walPath, line, utf-8);  // always runs
this.activeIds.delete(id);                    // no-op for ghost IDs
```
For an ID that was never written, this emits a tombstone that references nothing — wasted I/O. An early guard would eliminate this:
```ts
if (!this.activeIds.has(id)) return;
```
Not a regression (old code did the same), but this PR is an opportunity to close it completely.

**2. Set desync if `fsyncAfterWrite()` throws in `remove()`**

The sequence in `remove()` is:
```ts
appendFileSync(...)      // tombstone on disk
this.activeIds.delete(id) // Set updated ← can't be rolled back
this.fsyncAfterWrite()   // if this throws →
if (this.activeIds.size === 0) this.clear()  // never reached
```
If `fsyncAfterWrite` throws, the ID is deleted from the Set but compaction may not run. The catch block fires `capturePluginError` but the Set is now ahead of disk. This is recoverable after a restart (constructor re-seeds from disk), but a live instance would believe the WAL is empty when it isn't. This scenario is unlikely in practice but worth documenting.

**3. No test for constructor try-catch path**

The `try { this.activeIds = new Set(readAll()...) } catch { this.activeIds = new Set() }` branch is untested. A test with a path pointing to a directory (as the comment mentions) would solidify this.

### 🔍 Verdict

**Approve with minor suggestions.** The core fix is correct, well-tested for the happy path, and delivers the promised O(n²) → O(n) improvement for batch deletes. The ghost-tombstone issue is a missed optimization, not a regression. The fsync edge case is low-probability but worth a code comment. Overall solid work.

Signed: 🤖 GPT Council Member

PR #619, opened 2026-03-18
fix: ProposalsDB and IssueStore use untyped 'as any' row casts instead of typed query (#596)

### LINKED ISSUE #596: ProposalsDB and IssueStore use untyped 'as any' row casts instead of typed query results
## Problem

`ProposalsDB` and `IssueStore` still cast SQLite rows to `any`, which loses compile-time validation of column names and row shape.

### Exact code locations

`extensions/memory-hybrid/backends/proposals-db.ts`
- **129**: `get(...) as any`
- **149**: `.all(...params) as any[]`
- **169**: count row cast `as any`

`extensions/memory-hybrid/backends/issue-store.ts`
- **81**: `get(...) as any`
- **197**: `.all(...params) as any[]`
- **218**: `.all(term, term) as any[]`

Examples from current code:

```ts
// proposals-db.ts:129
const row = this.db.prepare("SELECT * FROM proposals WHERE id = ?").get(id) as any;

// issue-store.ts:81
const row = this.db.prepare("SELECT * FROM issues WHERE id = ?").get(id) as any;
```

That means downstream mapping functions accept effectively untyped data.

## Suggested fix

Define explicit row interfaces and cast to those instead:

```ts
interface ProposalRow {
  id: string;
  type: string;
  status: string;
  name: string;
  content: string;
  created_at: number;
  updated_at: number;
}

const row = this.db.prepare("SELECT * FROM proposals WHERE id = ?").get(id) as ProposalRow | undefined;
```

Do the same for `IssueRow` and update `rowToEntry` / `rowToIssue` signatures accordingly.

## Impact

- Column/property typos will compile but fail at runtime
- Database schema expectations are implicit instead of documented in types
- Makes future refactors riskier for relatively little gain

## Acceptance criteria

- `as any` row casts are removed from both backends
- Row mapping helpers accept typed row interfaces
- TypeScript catches invalid row property access in these modules

markus-lassfolk:
## Summary

- Adds `ProposalRow` and `CountRow` interfaces to `proposals-db.ts` documenting the exact SQLite column layout
- Adds `IssueRow` interface to `issue-store.ts` with fully typed columns including `IssueSeverity`
- Replaces all `as any` and `as any[]` row casts with typed double-casts (`as unknown as ProposalRow`, etc.) and updates `rowToEntry`/`rowToIssue` signatures to accept the typed row interfaces

## Test plan

- [x] All 3618 tests pass (`npm test`)
- [x] TypeScript type check clean (`tsc --noEmit`)
- [x] No `as any` row casts remain in either backend
- [x] Column/property typos in row mapping will now be caught at compile time

Closes #596

---

> [!NOTE]
> **Low Risk**
> Low risk: changes are TypeScript typing-only (replacing `as any` with explicit row interfaces and casts) with no SQL, schema, or runtime behavior changes expected.
> 
> **Overview**
> Tightens typing around SQLite query results in `IssueStore` and `ProposalsDB` by introducing `IssueRow`/`ProposalRow` (plus `CountRow`) and updating `rowToIssue`/`rowToEntry` to consume these typed shapes.
> 
> All previous `as any`/`as any[]` result casts are replaced with `as unknown as ...` and `IssueStore` now explicitly casts `severity` to `IssueSeverity`, making column/property mismatches fail at compile time.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7be1964a33e7b7fc2b84260c70cfd5074b691c8b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/extensions/memory-hybrid/services/embeddings/openai-provider.ts
+++ b/extensions/memory-hybrid/services/embeddings/openai-provider.ts
@@ -129,12 +129,44 @@ export class Embeddings implements EmbeddingProvider {
   async embedBatch(texts: string[]): Promise<number[][]> {
     if (texts.length === 0) return [];
 
-    const allResults: number[][] = [];
-    for (let i = 0; i < texts.length; i += this.batchSize) {
-      const batch = texts.slice(i, i + this.batchSize);
+    // Phase 1: Prefill results from cache (same model-order logic as embed()).
+    // Check each model in preference order so a cached fallback result is reused
+    // even if the primary model's cache slot is empty (#589).
+    const results: (number[] | undefined)[] = new Array(texts.length).fill(undefined);
+    const uncachedIndices: number[] = [];
+    let cacheHitModel: string | undefined;
+    for (let i = 0; i < texts.length; i++) {
+      let found = false;
+      for (const model of this.models) {
+        const cacheKey = makeCacheKey(model, texts[i]);
+        const cached = this.cache.get(cacheKey);
+        if (cached !== undefined) {
+          // LRU refresh: move to end
+          this.cache.delete(cacheKey);
+          this.cache.set(cacheKey, cached);
+          results[i] = cached;
+          found = true;
+          if (cacheHitModel === undefined) cacheHitModel = model;
+          break;
+        }
+      }
+      if (!found) uncachedIndices.push(i);
+    }
+
+    if (uncachedIndices.length === 0) {
+      if (cacheHitModel !== undefined) this.modelName = cacheHitModel;
+      return results as number[][];
+    }
+
+    // Phase 2: Batch-embed only the uncached texts, chunked by batchSize.
+    const uncachedTexts = uncachedIndices.map((i) => texts[i]);
+    const freshVectors: number[][] = [];
+    for (let i = 0; i < uncachedTexts.length; i += this.batchSize) {
+      const batch = uncachedTexts.slice(i, i + this.batchSize);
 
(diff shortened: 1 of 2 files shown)


markus-lassfolk (commented):
## Code Council Review — PR #619: Typed row casts ✅

**IssueRow / ProposalRow interface completeness**

Both interfaces look complete against the fields accessed in `rowToIssue` / `rowToEntry`. All JSON-serialised columns (`symptoms`, `tags`, `metadata`, `related_facts`, `evidence_sessions`) are correctly typed as `string` (pre-parse). Nullable columns are properly `string | null` or `number | null`. The added `CountRow` for `countRecentProposals` is a nice touch.

**`as unknown as IssueRow` — safety and idiom**

The double-cast pattern is the idiomatic TypeScript way to assert a type for an opaque runtime value (SQLite rows returned as `unknown` from `node:sqlite`). It's a clear, intentional statement that the developer accepts responsibility for schema alignment — better than `as any`, which suppresses all downstream type-checking. Not a runtime validator (zod/io-ts), but that's a separate concern and acceptable here.

**IssueSeverity import**

Added as `import type { ..., IssueSeverity }` — correct. Used only as a type cast (`row.severity as IssueSeverity`), so the type-only import is appropriate and won't introduce runtime overhead.

**No real issues found.** ✅ Clean, self-consistent change that tightens the type surface without adding complexity.

PR #626, opened 2026-03-18
fix: init-databases.ts has cascading env var fallback chains that are hard to audit (#598)

### LINKED ISSUE #598: init-databases.ts has cascading env var fallback chains that are hard to audit
## Problem

`init-databases.ts` resolves provider API keys via several slightly different fallback chains, which makes credential precedence hard to audit.

### Exact code locations

Provider-specific resolution branches:

- Google: **`extensions/memory-hybrid/setup/init-databases.ts:422-430`**
- OpenAI: **`init-databases.ts:433-449`**
- Anthropic: **`init-databases.ts:453-469`**
- OpenRouter: **`init-databases.ts:490-514`**
- Generic fallback env lookup: **`init-databases.ts:543-559`**

Examples of the drift:

```ts
// Google: init-databases.ts:423-424
const apiKey =
  resolveApiKey(providerCfg?.apiKey ?? cfg.distill?.apiKey) ??
  (process.env.GOOGLE_API_KEY?.trim() || undefined);

// OpenAI: init-databases.ts:439-443
const apiKey =
  resolveApiKey(providerCfg?.apiKey) ??
  (hasCustomExternalBaseURL ? undefined : gatewayToken) ??
  (hasCustomExternalBaseURL ? undefined : cfg.embedding.apiKey) ??
  (process.env.OPENAI_API_KEY?.trim() || undefined);
```

So the effective precedence depends on provider-specific handwritten logic instead of a shared resolver.

## Suggested fix

Extract a single `resolveProviderApiKey(provider, providerCfg, cfg, env)` helper that returns both:

- the resolved key
- metadata about **which source won** (`config`, `distill.apiKey`, `gatewayToken`, `OPENAI_API_KEY`, etc.)

Example shape:

```ts
type ResolvedApiKey = { value?: string; source?: string };
function resolveProviderApiKey(provider: string, ...): ResolvedApiKey { ... }
```

Also document the precedence rules in one table in the docs.

## Impact

- Hard to reason about which credential will actually be used
- Easy for provider behavior to drift over time
- Debugging misconfiguration is slower because the resolution path is not centralized or logged

## Acceptance criteria

- API key resolution is centralized in one helper
- Provider-specific exceptions are explicit and documented
- Debug logging can report the chosen source without exposing the secret value
- Docs describe precedence for each supported provider

markus-lassfolk:
## Summary

- Extracts a single `resolveProviderApiKey(prefix, providerCfg, cfg, resolveKey, opts)` helper that centralises all API-key resolution for every built-in provider
- Introduces `ResolvedApiKey = { value?: string; source: string }` so the winning source can be inspected without exposing the key itself
- Replaces ~120 lines of duplicate inline fallback chains in `resolveClient` with clean single-call delegations
- Documents the full precedence table in JSDoc (one authoritative source of truth)
- Injects `env` parameter for hermetic unit tests — no live `process.env` access

## Test plan

- [x] 26 new unit tests in `provider-routing.test.ts` covering every provider's fallback chain and `source` metadata
- [x] All 141 test files / 3638 tests pass (`vitest run`)
- [x] No TypeScript errors in changed files (`tsc --noEmit`)

Closes #598

---

> [!NOTE]
> **Medium Risk**
> Changes the API-key selection logic used for routing LLM requests across providers, including OpenAI gateway-token vs external baseURL handling; mistakes could break provider auth or accidentally choose the wrong credential source. Added targeted unit coverage reduces risk but this still touches core routing/auth paths.
> 
> **Overview**
> **Centralizes and audits provider API-key resolution** by introducing `resolveProviderApiKey()` (returning `{ value, source }`) and documenting a single precedence table for built-in and custom providers.
> 
> Updates `resolveClient()`/OAuth routing checks to delegate all key selection (including `<PREFIX>_API_KEY` env fallbacks) to this helper, with an explicit security guard to *avoid using* `gatewayToken`/`embedding.apiKey` when `openai` is configured with a custom external `baseURL`.
> 
> Adds extensive unit tests in `provider-routing.test.ts` to validate key precedence, env injection for hermetic tests, and correct `source` metadata across providers.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 05300384aa88ea30db5ac56d8f6466e550d3ecf5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/extensions/memory-hybrid/setup/init-databases.ts
+++ b/extensions/memory-hybrid/setup/init-databases.ts
@@ -582,7 +582,7 @@ function buildMultiProviderOpenAI(
       const { value: apiKey } = resolveProviderApiKey("ollama", providerCfg, cfg, resolveApiKey);
       const cacheKey = `ollama:${baseURL}`;
       return {
-        client: getOrCreate(cacheKey, () => new OpenAI({ apiKey: apiKey!, baseURL })),
+        client: getOrCreate(cacheKey, () => new OpenAI({ apiKey: apiKey ?? "ollama", baseURL })),
         bareModel,
         ollamaBaseUrl,
       };


markus-lassfolk (commented):
## Code Council Review — PR #626: Env var resolver ✅ with one minor note

**resolveProviderApiKey correctness**

The function correctly implements all provider-specific chains and the precedence table in the JSDoc matches the code. Each branch is explicit and easy to audit — this is a clear improvement over the scattered inline chains.

**ollama non-null assertion (`apiKey!`)**

The `!` on line `new OpenAI({ apiKey: apiKey! })` is safe. The `resolveProviderApiKey` for `"ollama"` always returns `{ value: "ollama", ... }` — the `value` field is always defined for this branch. TypeScript can't infer this from the return type (`ResolvedApiKey` allows `value?: string`), so the assertion is the right way to handle it. Alternatively, a local fallback could be used: `apiKey ?? "ollama"`, which would avoid the assertion without any behavior change — but this is a nit, not a bug.

**Behavior parity with the previous inline chains**

- Google: matches (`providerCfg.apiKey → distill.apiKey → GOOGLE_API_KEY`)
- OpenAI: matches, including the `hasCustomExternalBaseURL` guard on `gatewayToken` and `embedding.apiKey`
- Anthropic / OpenRouter / MiniMax: matches (single env fallback only)
- Ollama: matches (`providerCfg.apiKey ?? "ollama"`)
- Custom/unknown: **behavior changes slightly** — the old code had a separate block that only triggered if `providerCfg?.apiKey || providerCfg?.baseURL` was set, falling through to a second env-key block. The new code merges these into a single resolver call. The net result is identical behavior: `resolvedApiKey ?? "no-key"` is used if `providerCfg?.baseURL || resolvedApiKey` is truthy. No functional regression detected.

**Generic env fallback (`<PREFIX>_API_KEY`)**

Edge cases considered:
- Empty string: `.trim()` guard in the resolver ensures an empty env var is treated as absent. ✅
- Provider prefix is already uppercase in env: The resolver does `prefix.toUpperCase()`, so a prefix passed as `"myprovider"` correctly maps to `MYPROVIDER_API_KEY`. ✅
- Prefix with hyphens/special chars: If a provider name like `"my-provider"` is used, `toUpperCase()` alone won't convert the hyphen — env key would be `MY-PROVIDER_API_KEY`, which is not a valid env var name on most shells. This was true in the old code too, so **not a regression**, but worth noting as a pre-existing edge case.

**Test coverage**

The new test suite in `provider-routing.test.ts` is comprehensive and correctly covers all branches, including the `hasCustomExternalBaseURL` security guard and the `source` metadata field.

**Verdict: ✅ Approve.** The refactor achieves its stated goal — auditable, centralized key resolution — with no behavioral regressions. The ollama `!" assertion is safe; the hyphened-prefix edge case is pre-existing and out of scope.

PR #627, opened 2026-03-18
fix: feat: Humanizer style scoring as a quality-loop metric in hybrid-memory (#616)

### LINKED ISSUE #616: feat: Humanizer style scoring as a quality-loop metric in hybrid-memory
## Problem

There's no automated way to measure whether agent replies have drifted toward robotic/canned AI writing patterns. Persona instructions tell the model what to do, but there's no feedback signal for when output still sounds like a chatbot.

## What Humanizer is

Humanizer is a local Node CLI/library (ClawHub skill) that scores text against 29 AI-writing patterns and 500+ vocabulary terms.

- Install: `clawhub install humanizer`
- GitHub: https://github.com/brandonwise/humanizer
- Runs offline, no API key, Node >=18

Scoring breakdown:
- Pattern score (70%): explicit phrase/style detectors, vocab tiers, breadth bonuses
- Uniformity score (30%): burstiness, type-token ratio, sentence-length variation, trigram repetition

### Phase 1: Quality loop metric (spike — evaluator only, no rewriting)
Add a `humanizer_score` evaluation pass to the hybrid-memory quality loop:

1. After agent reply is sent, run: `humanizer score --json "<reply_text>"`
2. Parse output: `{ score, patterns_triggered, category_breakdown }`
3. Store in hybrid-memory:
   ```
   memory_store({
     category: "quality_loop",
     text: "humanizer_score: 0.73, patterns: ['great_question', 'happy_to_help'], model: sonnet, skill: weather"
   })
   ```
4. Periodic `memory_reflect` will surface patterns like "Sonnet scores high on gush patterns when using weather skill"

### Phase 2: Threshold-based selective rewrite
Only if score > configurable threshold (e.g. 0.7):
- Trigger single constrained rewrite prompt: "Keep meaning identical. Remove canned assistant phrasing. Do not add fluff or new content."
- Re-score and log delta
- Gate: prose-only channels (WhatsApp, email). Skip for: code explanations, incident reports, structured output.

### Phase 3: Batch scoring for generated docs
Run Humanizer on generated content before publishing:
- Release notes
- Issue descriptions (Scholar-generated)
- PR summaries
- Email drafts

## Acceptance criteria

- [ ] `humanizer score` CLI working on sample agent output
- [ ] Score + triggered patterns stored in `quality_loop` memory category
- [ ] Aggregate stats queryable: worst patterns by model/skill
- [ ] Threshold-based rewrite tested on ≥10 real agent replies and evaluated
- [ ] Config flag: `quality.humanizer.enabled` (default: false until validated)

## Files affected

- `extensions/memory-hybrid/lifecycle/hooks.ts` — add post-reply scoring hook
- `extensions/memory-hybrid/config/types/` — quality config options
- `skills/humanizer/SKILL.md` (new)

markus-lassfolk:
## Summary

Closes #616

Implements **Phase 1** of Humanizer style scoring — an offline quality-loop metric for detecting AI-writing patterns in agent replies. No rewriting; evaluator only.

- **`services/humanizer-score.ts`** — Core service with pure, testable functions:
  - `runHumanizerScore(text, cfg)` — runs `humanizer score --json` as a subprocess; returns null gracefully on ENOENT (not installed) or short text
  - `parseHumanizerOutput(raw)` — validates and normalizes the JSON output `{ score, patterns_triggered, category_breakdown }`
  - `formatQualityLoopEntry(result, opts)` — produces compact, searchable memory text: `humanizer_score: 0.73, patterns: ['great_question'], model: sonnet, skill: weather`

- **`config/types/features.ts`** — New `HumanizerConfig` type (enabled, bin, minTextLength, maxTextLength, modelTag, skillTag)
- **`config/types/index.ts`** — `humanizer: HumanizerConfig` field added to `HybridMemoryConfig`
- **`config/parsers/features.ts`** + **`config/parsers/index.ts`** — `parseHumanizerConfig` with safe defaults; opt-in (`enabled: false` by default)

- **`lifecycle/stage-capture.ts`** — New step 2 after frustration capture: scores last assistant message, stores result with `category: "quality_loop"` so `memory_reflect` can surface patterns like *"Sonnet scores high on gush patterns when using weather skill"*

- **`tests/humanizer-score.test.ts`** — 28 tests covering all pure functions + config parser

## 8-Point Blueprint alignment

- **Engineering Goal 1 (Stability)**: Errors are captured via `capturePluginError`; ENOENT (humanizer not installed) returns null silently
- **Engineering Goal 3 (Separation of Concerns)**: Scoring, storage, and reflection are cleanly separated — new service handles only scoring
- **Engineering Goal 4 (Testability)**: Pure functions `parseHumanizerOutput` and `formatQualityLoopEntry` are tested in isolation without any external dependencies

## Test plan

- [x] All 142 existing test files pass (3638 tests)
- [x] 28 new tests for humanizer service all pass
- [x] `npx tsc --noEmit` — clean
- [x] Feature is opt-in (`humanizer.enabled = false` by default) — no impact on existing deployments
- [x] Graceful degradation when `humanizer` binary is not installed

---

> [!NOTE]
> **Medium Risk**
> Adds a new opt-in capture-stage subprocess call (`humanizer` CLI) and persists its output to the facts DB, which could impact session-end latency/robustness if misconfigured or the binary behaves unexpectedly (though errors are handled and feature is disabled by default).
> 
> **Overview**
> Adds an **opt-in “Humanizer” style scoring** feature that runs at `agent_end` to evaluate the last assistant reply for AI-writing patterns and stores the result as a `quality_loop` fact.
> 
> Introduces `HumanizerConfig` (parser + config wiring) and a new `services/humanizer-score.ts` runner that executes `humanizer score --json` via stdin, parses/normalizes the output, and formats a compact stored entry; includes new unit tests covering parsing/formatting, config defaults, and graceful ENOENT/length handling.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 43866911aac9b4c2ab582d88d43d49b956455f23. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


markus-lassfolk (commented):
## Code Review — PR #627: Humanizer Style Scoring as Quality-Loop Metric

Overall this is a clean, well-structured addition. The feature is properly opt-in, the subprocess invocation is safe, config parsing is defensive, and graceful degradation is solid. A few issues worth addressing before merge.

---

### ✅ What's done well

**Subprocess security** — text is passed via `stdin`, not as a CLI argument, so content never appears in process listings or shell history. `spawn()` (not `exec()`) is used, eliminating shell-injection risk entirely. The `cfg.bin` path is taken from trusted user config, not from external input, which is the correct threat model.

**Graceful degradation** — ENOENT returns `null` silently ✓, non-zero exit codes are caught ✓, `child.stdin.on("error", () => {})` correctly prevents unhandled error events on broken pipe ✓, and the outer `catch` in `runCapture` logs at `info` severity so a missing binary never takes down a session.

**Config parsing** — the `minTextLength ≤ maxTextLength` constraint, the 20,000-char cap, the `=== true` strictness for `enabled`, and the whitespace trimming on `bin`/`modelTag`/`skillTag` are all exactly right. The refactor that pulls `extractLastAssistantText` into its own function is a genuine improvement to the existing code.

---

### 🔴 Issues

**1. Missing `await` on `ctx.factsDb.store()` (potential silent data loss)**

```ts
// lifecycle/stage-capture.ts ~line 120
ctx.factsDb.store({
  text: entryText,
  category: "quality_loop",
  ...
});
```

If `factsDb.store()` is async (Promise-returning) and the session teardown path in `runCapture` proceeds to `clearSessionState()` before the write resolves, the quality-loop fact may never be persisted. All sibling `factsDb` call-sites in the file should be checked — if they `await`, this one must too. If `store()` is intentionally synchronous, a comment to that effect would prevent this question arising again.

**2. Hardcoded 10 s timeout not in config**

```ts
const child = spawn(cfg.bin, ["score", "--json", "--stdin"], {
  timeout: 10_000,   // ← hardcoded
});
```

Slow CI boxes or cold-start containers may legitimately need more than 10 s. `HumanizerConfig` should expose a `timeoutMs` field (default: `10_000`) so operators can tune it. The parser already has a clean pattern for adding numeric config fields.

**3. `modelTag`/`skillTag` injection into fact text**

`formatQualityLoopEntry` concatenates tags directly into the fact string using `, ` as a delimiter:

```
humanizer_score: 0.73, patterns: ['gush'], model: sonnet, skill: weather
```

A `modelTag` value of `"sonnet, patterns: ['injected']"` would produce a spoofed fact that misleads `memory_reflect` queries. These values come from user config (trusted), but a one-line sanitise — strip commas and leading/trailing whitespace — would make the format robust without adding complexity. Alternatively, adopt a structured format (JSON or `key=value` pairs) for the stored fact.

---

### 🟡 Observations (non-blocking)

**4. `runHumanizerScore` test coverage gaps**

The happy path (successful subprocess invocation returning valid JSON) is not tested. Neither are: non-zero exit code handling, the 1 MB stdout guard, or the exact-boundary case for `minTextLength`. These paths hold the most interesting logic. A mock-subprocess approach (writing a tiny shell script that echoes valid JSON) would cover them without depending on the real binary.

**5. `rawOutput` naming is slightly misleading**

The `rawOutput` field in `HumanizerResult` is reconstructed from the parsed fields, not from the original JSON string. It's effectively a "lossless round-trip" of what was parsed, not the raw subprocess output. Renaming to `parsedFields` or `sourceData` avoids the impression that it holds the verbatim string.

**6. `stderr` silently accumulated, never logged**

```ts
child.stderr.on("data", (chunk) => { stderr += chunk; });
```

`stderr` is captured but only used in the rejection message for non-zero exit codes. If the process exits with code 0 but wrote warnings to stderr (valid in some humanizer builds), that context is dropped. Piping it through `api.logger.debug` on successful exit would help diagnose unexpected scores.

---

### Summary

| Severity | Item |
|---|---|
| 🔴 High | Missing `await` on `factsDb.store()` — verify or fix |
| 🔴 Medium | Hardcoded 10 s timeout should be configurable |
| 🟡 Low | `modelTag`/`skillTag` delimiter-injection in fact text |
| 🟡 Low | `runHumanizerScore` happy-path tests absent |
| ℹ️ Nitpick | `rawOutput` naming; stderr not logged on success |

The ENOENT graceful return, the stdin-based text passing, the `minTextLength` gate, and the overall refactor quality are all strong. Fix the `await` question and the timeout configurability, and this is ready to merge.

— Gemini Council

markus-lassfolk (commented):
## Security & Architecture Review — PR #627

Overall this is a solid, well-scoped Phase 1. The opt-in default, stdin-pipe design, ENOENT grace, and stdout size guard are all correct instincts. A few issues need addressing before merge.

---

### 🔴 High — `ctx.factsDb.store()` is not awaited

```ts
ctx.factsDb.store({
  text: entryText,
  category: "quality_loop",
  ...
});
```

There is no `await` here. If `factsDb.store` is async (even partially — e.g. writes to an async queue), the fact may be silently dropped because `clearSessionState` runs synchronously immediately after in step 4. At minimum add `await` or explicitly document that `store` is synchronous-safe. A silent fact-drop on every session is worse than a thrown error.

---

### 🔴 High — Architecture: blocking session_end for up to 10 seconds

`runHumanizerScore` is awaited inline in the critical path of `runCapture`. With a 10-second subprocess timeout, every session close can stall for 10 seconds before session state is cleaned up, event logs are written, and compaction runs (steps 3–6). Even though session_end is not latency-sensitive for the user, it blocks the full cleanup chain.

**Recommendation:** either run the scoring in a detached `setImmediate`/`Promise.resolve().then()` after session cleanup, or cap the timeout tighter (2–3s) and document the trade-off. A dedicated async queue is overkill for Phase 1, but don't gate cleanup on it.

---

### 🟡 Medium — Timeout kill path produces useless error message

When `spawn()`'s 10s timeout fires, the child is SIGTERMed and the `close` handler receives `code=null, signal='SIGTERM'`. The check `if (code === 0)` fails → `reject(new Error(`humanizer exited with code null: ${stderr}`))`. At timeout, `stderr` is typically empty, so the error logged is `"humanizer exited with code null: "` — completely opaque. Suggest catching SIGTERM explicitly:

```ts
child.on("close", (code, signal) => {
  if (signal) return reject(new Error(`humanizer killed by signal ${signal}`));
  if (code === 0) return resolve(parseHumanizerOutput(stdout));
  reject(new Error(`humanizer exited with code ${code}: ${stderr}`));
});
```

---

### 🟡 Medium — `formatQualityLoopEntry` format is fragile for downstream parsing

The stored fact is a free-text string: `"humanizer_score: 0.73, patterns: ['great_question', 'happy_to_help'], model: sonnet"`

Pattern names are embedded raw inside `[' ']` delimiters with no escaping. If the humanizer binary ever returns a pattern name containing `'`, `,`, or `]` (valid in some CLI tools), downstream parsing in `memory_reflect` will break silently. Either:
- Store as a structured JSON value field (key: `humanizer_score`, value: JSON blob) so it's machine-readable, or
- At minimum, sanitize pattern strings before embedding: `p.replace(/'/g, '')`

Also: no session identifier or ISO timestamp is embedded. When `memory_reflect` analyzes 200 accumulated facts, it cannot order them chronologically or correlate them to a session. Recommend adding at minimum `ts: ${Date.now()}` to the stored text.

---

### 🟡 Medium — `modelTag` / `skillTag` are static config, not dynamic

These tags are set once in config and applied to every fact. If a deployment runs multiple models or skills in the same instance, every quality_loop fact gets the same static tag regardless of what model actually produced the reply. Consider whether these should be injected from the lifecycle context (e.g. `ctx.cfg.model` or session metadata) rather than being static strings.

---

### 🟢 Low — `child.stdin.on("error", () => {})` is correct but undocumented

Silencing EPIPE on stdin is correct behavior (process exited before we finished writing), but a one-line comment explaining why would prevent future reviewers from removing it as dead code.

---

### 🟢 Low — `decayClass: "normal"` (~2w half-life) may be too short

Quality-loop patterns are only meaningful with enough statistical samples. A 2-week decay means early scores are gone before `memory_reflect` has seen enough data to surface real trends, especially for low-traffic agents. Consider `"durable"` (~3mo) or making it configurable.

---

### ✅ What's done right

- **stdin pipe** over CLI args — no injection via message content
- **spawn, not exec** — no shell involved, `bin` path injection is not exploitable
- **ENOENT → null** — graceful on missing binary, no error surface to users
- **stdout 1MB cap** — good defense against runaway output
- **opt-in by default** — correct for a new metric
- **28 tests on pure functions** — clean separation of testable logic from I/O
- **maxTextLength=4000 cap** — avoids leaking entire long replies to an external process

---

Blocking on: (1) the missing `await` on `factsDb.store` and (2) the session cleanup sequencing. The timeout signal handling and format fragility are strong-recommend fixes before this lands in production.

— GPT Council

Issue #1625, opened 2026-05-24
Active-task ledger should refresh from live issue/PR state during normal stewardship


markus-lassfolk:
## Problem

The active-task ledger can remain stale until a later deep-verification pass, even when live state is already clear from GitHub or another authoritative source.

This showed up during the 2026-05-24 task cleanup: normal agents repeatedly checked live GitHub state, correctly concluded that the relevant issues and PRs were closed/merged, but the task tracking view did not reliably converge to that state until explicit cleanup/deep verification happened.

Issue #1624 covers one concrete renderer bug where equal-timestamp facts can preserve stale `status=in_progress`. This issue tracks the broader workflow gap: state capture should be live and incremental enough that we do not need to wait for a deep verifier to refresh the active-task list.

## Expected behaviour

When a task references live external state, normal stewardship should be able to refresh and reconcile that state into the task ledger promptly.

Examples:

- a task tied to a GitHub issue should notice when the issue is closed
- a task tied to a PR should notice when the PR is merged/closed
- a goal/task whose acceptance criteria are backed by live repo state should update `status`, `next`, and `task_updated` based on that evidence
- terminal live state should clear active rows without waiting for a periodic deep verification sweep

## Why this matters

Waiting for deep verification creates noisy loops:

1. an agent checks GitHub and sees no open work
2. the local active-task projection still says work is active
3. a later agent repeats the same audit
4. the system spends cycles proving the same terminal state again

The user-facing effect is that Maeve appears to know the correct answer but fails to keep her own tracking clean. That erodes trust in the task ledger.

## Proposed direction

Add a normal "live-state reconciliation" path for active tasks/goals.

Possible implementation shape:

- Store structured external references on project facts, e.g. `github_repo`, `issue_number`, `pr_number`, `verification_target`, or a typed provenance object.
- During heartbeat/stewardship/task projection generation, fetch the current state for referenced issues/PRs within a small budget.
- If the live source is terminal, checkpoint the task as `done` / `closed` with evidence in `next` or provenance.
- If the live source is still active, refresh `task_updated` and keep the next action concrete.
- Apply deterministic conflict handling so live terminal evidence wins over stale non-terminal local facts.
- Keep deep verification for broad audits and high-risk final sign-off, not basic state convergence.

## Acceptance criteria

- Active tasks that reference a closed GitHub issue or merged/closed PR are automatically updated during normal stewardship, not only during deep verification.
[...]

PR #1626, opened 2026-05-24
Active-task renderer can mix stale status with newer task fields when facts share timestamp

### LINKED ISSUE #1624: Active-task renderer can mix stale status with newer task fields when facts share timestamp
## Problem

The hybrid-memory active-task projection can show a stale task as still active even after repeated canonical updates mark it closed/done.

Observed during a live cleanup on 2026-05-24: agents repeatedly checked live GitHub state for `markus-lassfolk/openclaw-hybrid-memory` and correctly concluded that the relevant issues and PRs were closed/merged. The active-task/projection state still kept stale rows around, producing a split-brain task display such as:

- old `status = in_progress`
- newer `next = Closed...` / cleanup text from a later checkpoint

That makes the task ledger look unresolved even when the task has already reached a terminal state.

## Current evidence

On `main`, `extensions/memory-hybrid/services/task-ledger/canonical.ts` groups project facts by canonical entity and key, then keeps the latest fact only when:

```ts
if (!prev || f.createdAt > prev.createdAt) {
  km.set(k, f);
}
```

This has no deterministic tie-break when multiple project facts for the same entity/key land in the same timestamp bucket. If `status=in_progress` and a terminal `status=done` / `closed` update have equal `createdAt` precision, whichever fact is encountered first can win for that key. Other keys, such as `next`, can come from the later checkpoint, so the renderer can compose one task row from mutually inconsistent fact versions.

## Expected behaviour

Active-task rendering should select the newest fact deterministically for every `entity + key`, even when timestamps are equal or coarsely rounded.

A terminal status update should not be lost behind an earlier non-terminal status fact solely because both facts share the same `createdAt` value.

## Why this matters

This causes repeated stewardship churn:

- agents check live state and correctly conclude the work is done
- they checkpoint/clear the task
- the generated active-task projection still shows stale active work
- later agents repeat the same audit instead of moving on

This is operationally expensive and makes the task ledger untrustworthy.

## Suggested fix

Add a stable ordering/tie-break in `groupProjectFactsByEntity()`, for example:

1. compare `createdAt`
2. when equal, compare a monotonic/high-resolution field if available
3. otherwise tie-break by stable insertion/order/id/row id so the result is deterministic and newer writes win

If the `MemoryEntry` type does not expose a suitable monotonic field, the ledger should preserve query order explicitly or the store/query should return enough metadata to distinguish same-second/same-millisecond writes.

## Regression test

Add a test that creates multiple project facts for the same canonical task where at least two `status` facts share the same `createdAt`:

1. `status = in_progress`
2. `next = Closed by live GitHub audit`
3. `status = done` or `closed`

The active-task projection should render the task as terminal/done and should not keep it in the active list because of the earlier `in_progress` status.

Also test that mixed keys do not combine stale status with newer non-status fields in a way that presents a closed task as active.

## Related live observation

During the 2026-05-24 audit, the repo had no open PRs and the relevant PRs/issues had been closed or merged, but active-task state repeatedly retained stale rows until manual canonical checkpoint cleanup. This issue captures the underlying renderer/storage selection bug so it can be fixed in code rather than repeatedly cleaned up operationally.

markus-lassfolk:
## Fix Summary

When  encounters two facts for the same entity+key with identical  timestamps, it previously kept whichever was processed first — a non-deterministic outcome. This caused the active-task renderer to display stale  alongside newer  fields from the same checkpoint.

### Changes

****

Added a  helper called when  values are equal:

1. **For the  key**: a terminal status value (, , , , ) wins over a non-terminal one. This prevents a stale  from persisting when a terminal checkpoint lands in the same timestamp bucket.

2. **For all other keys**: lexicographic id comparison provides a stable total order so the result is deterministic regardless of iteration order.

### Test coverage

A regression test covers the specific scenario from #1624: two  facts with equal  where one is  and the other is  — the terminal fact must win.

Closes #1624

---

> [!NOTE]
> **Medium Risk**
> Changes how the “latest” fact is selected when `createdAt` ties, which can affect persisted/derived task status and display; risk is moderate but localized to task-ledger fact grouping.
> 
> **Overview**
> Fixes nondeterministic fact selection in `groupProjectFactsByEntity` when multiple facts share the same `createdAt`.
> 
> Adds a deterministic tie-break (`factTieBreak`): for `status`, terminal values win over non-terminal on equal timestamps; otherwise the fact with the lexicographically newer `id` wins to ensure stable ordering.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5c7f9434d65c20afc983b2f232851bc38dd5d14c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/extensions/memory-hybrid/services/task-ledger/canonical.ts
+++ b/extensions/memory-hybrid/services/task-ledger/canonical.ts
@@ -47,9 +47,39 @@ export function activeTaskProvenance(canonical: string, existing?: string | null
   return JSON.stringify(base);
 }
 
+/** Deterministic tie-break for two facts with equal createdAt.
+ *
+ *  Rules:
+ *  1. For the `status` key, terminal status wins over non-terminal when timestamps are equal.
+ *     This prevents a stale "in_progress" from persisting when a terminal "done"/"closed" update
+ *     lands in the same second/millisecond bucket.
+ *  2. In all other cases, newer id wins (lexicographic: later ids are "greater").
+ *     This gives a stable total order without depending on insertion/iteration order.
+ *
+ *  Returns a positive number if `b" is preferred (b wins), negative if `a` wins, 0 if identical.
+ */
+function factTieBreak(a: MemoryEntry, b: MemoryEntry, key: string): number {
+  if (a.createdAt === b.createdAt) {
+    // Rule 1: status terminal-vs-nonterminal tie-break
+    if (key === "status") {
+      const aTerminal = isTerminalFactStatus(a.value ?? "") ? 1 : 0;
+      const bTerminal = isTerminalFactStatus(b.value ?? "") ? 1 : 0;
+      if (aTerminal !== bTerminal) return bTerminal - aTerminal; // terminal (1) wins
+    }
+    // Rule 2: lexicographic id tie-break (stable, later id = newer fact)
+    return b.id.localeCompare(a.id);
+  }
+  return 0; // caller already compared createdAt; this is only for equal timestamps
+}
+
 /** Latest value per entity+key from non-superseded project facts.
  *  Entity labels are normalized via factCanonicalLabel() (provenance-aware + trim + toLowerCase + suffix/separator cleanup)
- *  so that case-variant entries (e.g. "Humanizer" / "humanizer") are merged into one group. */
+ *  so that case-variant entries (e.g. "Humanizer" / "humanizer") are merged into one group.
+ *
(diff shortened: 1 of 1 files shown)


markus-lassfolk:
This PR is a draft and is superseded by the canonical PR for the same issues: #1628 (non-draft, fix: active-task state convergence for #1624 and #1625). Closing this duplicate. @markus-lassfolk please close.

PR #1943, opened 2026-06-25
fix(maintenance): stop degraded contradiction backlog from aborting nightly cron (#1942)

### LINKED ISSUE #1942: bug: resolve-contradictions degraded backlog aborts the entire maintenance-nightly run; cron guard never updates
## Summary

When the autonomous contradiction resolver (`resolve-contradictions` step inside `openclaw hybrid-mem maintenance nightly`) crosses the degraded-backlog threshold (200 ambiguous pairs × 3 consecutive runs), it **throws an unhandled error out of the step runner** which causes `computeExitCode` to return `1`, the orchestrator to exit non-zero, and the nightly `maintenance-nightly` cron job to fail every single night thereafter.

The bug is **silent** to the operator: it looks like "the plugin failed", but it is in fact a structural over-reaction to a monitoring signal. The ambiguity backlog is a known and already-tracked operational concern (issues #1805, #1687, #1692, #1781) — those fixes make *progress* on the backlog. This bug is orthogonal: **even after those fixes ship**, the orchestrator will keep killing the entire nightly run the moment the backlog is unhealthy, because the abort gate is hard-coded into the step runner, not the autonomous resolution path.

## Impact

- `maintenance-nightly` cron job has been failing every night since **2026-06-23** (with `consecutiveErrors` climbing). The 20h agent-level guard file (`~/.openclaw/cron/guard/maintenance-nightly.ms`) is correctly *not* updated, so the next-run slot fires again, fails again, and the cycle repeats.
- All subsequent nightly steps after `resolve-contradictions` (`enrich-entities`, `extract-implicit`, `entity-mentions-cleanup`, `dream-cycle-core`, `continuous-verification`, `closed-loop-analysis`, `cross-agent-learning`, `tool-effectiveness`, `cost-log-prune`, `self-correction-run`, `reflect`, `extract-procedures`, `extract-directives`, `extract-reinforcement`, `repair-vectors`, `vectordb-optimize`, `scope-promote`, `implicit-feedback-collapse`, `audit-health`, `pending-digest`) are **never reached** once the gate trips. This is by design of `computeExitCode` — it inspects `results` after the run and decides the exit code from the whole batch — but the abort is being caused by a *single* step that should be emitting a degraded-state warning, not killing the orchestrator.
- `resolve-contradictions` itself still runs (it threw, so it produced a summary), so the operator gets a confusing mixed signal: the degraded-state line is right, the exit-code line is wrong.

## Reproduction

On an installation with an unresolved ambiguity backlog ≥ 200 for 3+ consecutive nightly runs:

```bash
# 1. Run the nightly maintenance — observe that resolve-contradictions throws
#    "resolve-contradictions degraded backlog (...)" and the orchestrator exits 1.
openclaw hybrid-mem maintenance nightly --verbose

# 2. Inspect the state file the gate uses
cat ~/.openclaw/cron/guard/resolve-contradictions-no-progress.json
# {
#   "consecutiveNoProgressRuns": 4,
#   "lastAmbiguous": 4223,
#   "lastAutoResolved": 0,
#   "updatedAt": "2026-06-25T02:23:48Z"
# }

# 3. Trigger the abort deterministically by overriding the thresholds low
openclaw hybrid-mem maintenance nightly --include resolve-contradictions \
  --degraded-ambiguous-threshold 1 --degraded-consecutive-threshold 1
# → exit 1
```

## Root cause

**File:** `dist/cli/commands/manage/maintenance-step-runners.js` lines 134-138

```js
function assertContradictionMaintenanceSummaryDoesNotBlock(summary, evaluation) {
    const semantic = parseSemanticTokenFromSummary(summary);
    if (evaluation.degraded || semanticOutcomeBlocksOrchestratorGuard(semantic))
        throw new Error(`resolve-contradictions degraded backlog (${summary})`);
    return summary;
}
```

The step runner for `resolve-contradictions` (line 276-286) calls this assertion, which **throws** when `evaluation.degraded === true` (set by `contradiction-progress-summary.js` line 36 when ambiguous ≥ 200 for ≥ 3 consecutive runs).

**Cascade:**

1. `contradiction-progress-summary.js` line 36 — `evaluateContradictionProgress` sets `evaluation.degraded = true` and `evaluation.exitCode = 2` when:
   - `ambiguous ≥ degradedAmbiguousThreshold` (default 200)
   - `autoResolved === 0` (no progress)
   - `consecutiveNoProgressRuns ≥ degradedConsecutiveThreshold` (default 3)
2. The auto-mode runner then **persists** the consecutive state to `~/.openclaw/cron/guard/resolve-contradictions-no-progress.json` so the counter survives across runs.
3. `maintenance-step-runners.js` line 136 — the assertion **throws an unhandled error**.
4. `maintenance-orchestrator.js` line 416 — `computeExitCode` returns 1 because `results.some(r => r.status === "failed")`.
5. `semantic-outcome.js` line 60 — `semanticOutcomeBlocksOrchestratorGuard("partial")` returns true (the runner sets `semantic=partial` in `runContradictionMaintenanceAutoStep`), so the orchestrator treats `partial` as a guard-blocking outcome.
6. The whole nightly run exits non-zero. The agent-level bash harness's `validate-cron-exit` correctly fails validation. The guard file at `~/.openclaw/cron/guard/maintenance-nightly.ms` is *not* updated. Next run, same thing.

## Evidence (operator-side)

| Date (UTC) | Run duration | `consecutive` | `ambiguous` | Outcome | Source log |
|---|---|---|---|---|---|
| 2026-06-13 | 11m18s | 0 | 78 | ok | `maintenance-nightly-20260613T020437Z-17945.log` |
| 2026-06-15 | 17m13s | 0 | 78 | ok | `maintenance-nightly-20260615T023412Z-15531.log` |
| 2026-06-23 | ~13m | (older plugin) | n/a | failed (`writeFileSync is not defined`) | `maintenance-nightly-20260623T023845Z-27522.log` |
| 2026-06-24 | ~51m | 3 | 3,693 | failed (degraded backlog) | `maintenance-nightly-20260624T020555Z-14931.log` |
| 2026-06-25 02:06 | 17m | 4 | 4,223 | failed (degraded backlog) | `maintenance-nightly-20260625T020609Z-27083.log` |
| 2026-06-25 02:23 | 18m | (post-throw, no summary printed before exit) | — | failed (orchestrator exit 1) | `maintenance-nightly-20260625T022301Z-17689.log` |

`validate-cron-exit` JSON on the 2026-06-25 02:23 run:
```json
{
  "maintenanceStatus": "failed",
  "failedSteps": [{
    "name": "maintenance-nightly",
    "exit": 1,
    "reason": "nonzero_exit",
    "error": "Non-zero exit code (nonzero_exit)"
  }],
  "guardUpdated": false
}
```

The `cron get hybrid-mem:maintenance-nightly` state shows `consecutiveErrors: 2` and rising.

The `resolve-contradictions-no-progress.json` state file (now exists on this installation):
```json
{
  "consecutiveNoProgressRuns": 4,
  "lastAmbiguous": 4223,
  "lastAutoResolved": 0,
  "updatedAt": "2026-06-25T02:23:48Z"
}
```

## Why this is a defect, not by-design

`computeExitCode` already has a distinct exit code for "degraded but acceptable" outcomes (`semantic=partial` → exit 1 only when it blocks the orchestrator guard). The intent in `contradiction-progress-summary.js` is clearly to surface a degraded-state *signal* so the operator can investigate. But the step runner conflates the signal with a fatal condition by throwing — and `semantic-outcome.js` line 60 promotes `partial` to guard-blocking regardless of the cause.

The gate logic in `contradiction-progress-summary.js` line 36 is doing its job: it correctly detects that the backlog is unhealthy and *records* the consecutive-runs counter. But the *response* (throw + abort) is the wrong one for a nightly cron job. Other steps in the same pipeline (`extract-implicit`, `continuous-verification`, `audit-health`, `pending-digest`) **also** report `failed` with `semantic=partial` and the orchestrator still records them as failed steps — but those failures don't throw at the runner layer; they return a summary and the orchestrator's `computeExitCode` reflects that as expected.

The `resolve-contradictions` step is the **only one** in `maintenance-step-runners.js` that wraps its step result in a `throw` for a degraded-state condition. That asymmetry is the bug.

## Suggested fix paths

1. **Minimum diff, recommended.** Change `assertContradictionMaintenanceSummaryDoesNotBlock` (line 134) so a degraded evaluation **returns** the summary with `semantic=partial` rather than throwing. Add an explicit `step.status = "degraded"` (a new step status) and update `computeExitCode` to treat `degraded` like `deferred` (exit 0 / non-blocking) but emit a structured warning to the cron validator. This keeps the operator-visible signal intact, lets the rest of nightly maintenance continue, and avoids the consecutive-runs counter going stale.

2. **Stronger: separate the operational signal from the run outcome.** Introduce a new step status `monitoring` (the `exitReason === "ambiguous_backlog_monitoring"` value already exists in `contradiction-progress-summary.js` line 30 but is unused). When evaluation is not `degraded` but `noProgress && thresholdExceeded`, return `status: "monitoring"` with `exitCode: 0`. The counter still increments, the operator still gets the signal, and the nightly run completes.

3. **Strongest: invert the default.** The current default makes the gate *trip* (consecutive-degraded → exit 1). Invert it so the gate only fires when the operator opts in (config flag `maintenance.resolveContradictions.failOnDegradedBacklog`, default false). The counter still records; the orchestrator just doesn't abort.

Any of the three preserves the signal. The current behavior — *silently killing nightly maintenance forever once the backlog is unhealthy* — is the worst of all worlds: the operator gets a permanent FAILED cron entry, the rest of the pipeline is starved, and the only way to recover is to manually drain the ambiguity backlog (which is exactly what the nightly maintenance is supposed to do).

## Related work (already merged or in flight)

- #1805, #1687, #1692, #1781 — making autonomous progress on the ambiguity backlog (autonomous resolution, LLM adjudication, ≥80% auto-resolve target). These are necessary but **not sufficient** to fix this bug: even at 100% auto-resolve rate, a single unhealthy run would still abort the orchestrator until the consecutive-runs counter resets.
- #1934 (closed 2026-06-24) — the previous `writeFileSync is not defined` bug in `cron-guard.ts` that masked this one. Different bug; not a duplicate.

## Environment

- Plugin: `openclaw-hybrid-memory@2026.6.241` (memory-manager 3.0)
- OpenClaw: `2026.6.8 (844f405)`
- facts.db: 64,120 facts; LanceDB vector index enabled; resolved-ambiguity backlog 4,223
- Affected cron job id: `hybrid-mem:maintenance-nightly` (schedule `5 2 * * *` UTC, isolated session, model `minimax/MiniMax-M3`)
- Affected state file: `~/.openclaw/cron/guard/resolve-contradictions-no-progress.json`
- Affected state file (agent-level guard): `~/.openclaw/cron/guard/maintenance-nightly.ms`

markus-lassfolk:
## Summary
- Fixes [#1942](https://github.com/markus-lassfolk/openclaw-hybrid-memory/issues/1942): `resolve-contradictions` no longer throws when the ambiguous backlog crosses the degraded threshold, so `maintenance-nightly` continues through remaining steps instead of exiting 1 every night.
- Introduces a non-blocking `semantic=monitoring` outcome for operational signals (degraded contradiction backlog and transient `record-storage-sample` unavailability) while preserving degraded telemetry in cron validation reports.
- Updates cron guard logic so monitoring-only issues (`resolve_contradictions_degraded_backlog`, `record_storage_unavailable`) surface in `reportableIssues` but no longer block guard file advancement when the run otherwise succeeds.
- Bumps plugin version to **2026.250**.

## Root cause
`assertContradictionMaintenanceSummaryDoesNotBlock` threw on `evaluation.degraded`, which marked the step failed and caused `computeExitCode` to return 1. Even returning without throwing would still have failed because `semantic=partial` blocks the orchestrator guard.

## Test plan
- [x] `npm test -- --run contradiction-progress-summary maintenance-orchestrator maintenance-job-run maintenance-validator-coverage cron-exit-validator`
- [ ] On an installation with degraded backlog, run `openclaw hybrid-mem maintenance nightly --verbose` and confirm subsequent steps execute after `resolve-contradictions`
- [ ] Confirm `validate-cron-exit` reports `maintenanceStatus=success`, `semanticStatus=degraded`, `guardUpdated=true` for degraded backlog with exit 0
- [ ] Confirm standalone `openclaw hybrid-mem resolve-contradictions --json` still exits degraded (`exitCode=2`) when thresholds trip

### DIFF
--- a/extensions/memory-hybrid/cli/commands/manage/maintenance-step-runners.ts
+++ b/extensions/memory-hybrid/cli/commands/manage/maintenance-step-runners.ts
@@ -72,6 +72,33 @@ function reflectRulesDiagnosticsIndicateFailure(
   return !d.parseSuccess || rulesStored === 0;
 }
 
+function formatReflectRulesMaintenanceSummary(r: {
+  rulesStored: number;
+  rulesExtracted: number;
+  diagnostics?:
+    | {
+        parseSuccess: boolean;
+        status: "ok" | "partial" | "degraded";
+        zeroRulesReason?: string;
+        modelResponseChars?: number;
+      }
+    | undefined;
+}): string {
+  const d = r.diagnostics;
+  const failed = reflectRulesDiagnosticsIndicateFailure(d, r.rulesStored);
+  return [
+    `rulesStored=${r.rulesStored}`,
+    `rulesExtracted=${r.rulesExtracted}`,
+    d ? `parse_success=${d.parseSuccess}` : null,
+    d?.zeroRulesReason ? `zero_rules_reason=${d.zeroRulesReason}` : null,
+    d ? `status=${d.status}` : null,
+    typeof d?.modelResponseChars === "number" ? `model_response_chars=${d.modelResponseChars}` : null,
+    failed ? "semantic=failed" : "semantic=success",
+  ]
+    .filter(Boolean)
+    .join(" ");
+}
+
 export interface BuildCliRunnersOptions {
   verbose?: boolean;
   backfillDecayMarker?: string;
@@ -457,19 +484,11 @@ export function buildCliMaintenanceRunners(
 
   set("reflect-rules", async () => {
     const r = await b.runReflectionRules({ dryRun: false, model: b.reflectionConfig.model, verbose });
-    const d = r.diagnostics;
-    const summary = [
-      `rulesStored=${r.rulesStored}`,
-      `rulesExtracted=${r.rulesExtracted}`,
-      d ? `parse_success=${d.parseSuccess}` : null,
-      d?.zeroRulesReason ? `zero_rules_reason=${d.zeroRulesReason}` : null,
-      d ? `status=${d.status}` : null,
-      d?.status === "degraded" || reflectRulesDiagnosticsIndicateFailure(d, r.rulesStored) ? "semantic=failed" : null,
-    ]
-      .filter(Boolean)
-      .join(" ");
-    if (d?.status === "degraded" || reflectRulesDiagnosticsIndicateFailure(d, r.rulesStored)) {
-      throw new Error(`reflect-rules semantic failure (${d?.zeroRulesReason ?? "degraded"}): ${summary}`);
+    const summary = formatReflectRulesMaintenanceSummary(r);
+    if (reflectRulesDiagnosticsIndicateFailure(r.diagnostics, r.rulesStored)) {
+      throw new Error(
+        `reflect-rules semantic failure (${r.diagnostics?.zeroRulesReason ?? r.diagnostics?.status ?? "failed"}): ${summary}`,
+      );
     }
     return summary;
   });
--- a/extensions/memory-hybrid/cli/commands/manage/maintenance-step-runners.ts
+++ b/extensions/memory-hybrid/cli/commands/manage/maintenance-step-runners.ts
@@ -210,25 +210,9 @@ function assertLifecycleSyncSummaryDoesNotBlock(summary: string): string {
   return summary;
 }
 
-function assertRecordStorageSampleSummaryDoesNotBlock(summary: string): string {
-  if (/\breason=storage_unavailable\b/i.test(summary)) {
-    const semanticSummary = summary.includes("semantic=") ? summary : `${summary} semantic=partial`;
-    throw new Error(`record-storage-sample storage unavailable (${semanticSummary})`);
-  }
-  return summary;
-}
-
 function resolveOpenclawHomeFromSqlitePath(resolvedSqlitePath: string | null | undefined): string {
   if (resolvedSqlitePath) return join(dirname(resolvedSqlitePath), "..");
   return join(homedir(), ".openclaw");
-}
-
-function assertContradictionMaintenanceSummaryDoesNotBlock(summary: string, evaluation: { degraded: boolean }): string {
-  const semantic = parseSemanticTokenFromSummary(summary);
-  if (evaluation.degraded || semanticOutcomeBlocksOrchestratorGuard(semantic)) {
-    throw new Error(`resolve-contradictions degraded backlog (${summary})`);
-  }
-  return summary;
 }
 
 export function buildCliMaintenanceRunners(
@@ -280,8 +264,7 @@ export function buildCliMaintenanceRunners(
       /* non-fatal */
     }
     const r = recordStorageGrowthSample(b.factsDb, lanceBytes, {});
-    const summary = `status=${r.status} reason=${r.reason ?? "none"} semantic=${r.reason === "storage_unavailable" ? "partial" : "success"}`;
-    return assertRecordStorageSampleSummaryDoesNotBlock(summary);
+    return `status=${r.status} reason=${r.reason ?? "none"} semantic=${r.reason === "storage_unavailable" ? "monitoring" : "success"}`;
   });
 
   set("analyze-maintenance-logs", async () => {
@@ -390,7 +373,7 @@ export function buildCliMaintenanceRunners(
 
   set("resolve-contradictions", async () => {
     const openclawHome = resolveOpenclawHomeFromSqlitePath(b.resolvedSqlitePath);
-    const { summary, evaluation } = await runContradictionMaintenanceAutoStep({
+    const { summary } = await runContradictionMaintenanceAutoStep({
       openclawHome,
       degradedAmbiguousThreshold: DEFAULT_AMBIGUOUS_BACKLOG_DEGRADED_THRESHOLD,
       degradedConsecutiveThreshold: ORCHESTRATOR_CONTRADICTION_DEGRADED_CONSECUTIVE_THRESHOLD,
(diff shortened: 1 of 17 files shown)


markus-lassfolk:
Addressed all three Codex review threads in 411b4d3 + latest push:

1. **Non-monitoring degraded summaries** — `shouldUpdateMaintenanceGuard` now takes `semanticStatus` and only allows guard advance on `degraded` when every reportable issue is in `MONITORING_ONLY_FAILURE_CLASSES`.

2. **Standalone CLI exit code** — restored shell `exitCode=2` for degraded backlog on standalone runs; cron wrappers (`HM_JOB` set) keep shell exit 0 so `hm_step`/`set -e` does not abort.

3. **Pre-threshold backlog runs** — `semantic=monitoring` is emitted only when `evaluation.degraded` is true (consecutive threshold reached), not during early `ambiguous_backlog_monitoring` runs.

Please resolve the review threads when satisfied.

PR #1951, opened 2026-06-25
fix: maintenance nightly exit codes + memory_store empty args (#1949, #1950)

### LINKED ISSUE #1949: maintenance nightly: 3 step failures produce misleading exit codes (extract-implicit maxWallClock, continuous-verification degraded, audit-health strict warnings)
## Summary

`openclaw hybrid-mem maintenance nightly` on a fully-installed host reports `Maintenance nightly — 36 steps (0 ok, 33 skipped, 0 deferred, 3 failed)` and exits with a non-zero code, even though the run completed successfully. The three "failures" are not real failures:

1. **`extract-implicit`** — hit `maxWallClockSeconds` (default 300s, configured 1200s). **128 signals were extracted** from 116 of 471 sessions and stored. The step is reported as `failed` because `partial=true`, even though the partial run was successful.
2. **`continuous-verification`** — `Machine status: status=degraded reason=errors_present checked=45 confirmed=0 stale=0 uncertain=45 errors=42`. All 45 facts went through the verifier; 42 surfaced errors during LLM calls and were marked `UNCERTAIN`. This is the verifier behaving correctly under degraded LLM conditions, but the step is reported as `failed`.
3. **`audit-health`** — `--strict` mode fails on **7 warnings** that are diagnostic / advisory, not errors:
   - personaProposals scopeFilter not set (advisory)
   - Unconfigured category `meta=3` (migration drift, 3 rows)
   - Implicit-feedback prefix histogram duplication (advisory, content-shape pattern)
   - Stop-word-like top entities `Credentials`, `User` (advisory)
   - 959 facts missing canonical embeddings (size of the gap)
   - Vector lifecycle SLO breach (3.64% actual vs 2.00% target — over by 1.64 percentage points)
   - 514 validated procedures not promoted (4 awaiting_approval, 510 low_recall)

The result: every nightly run reports failure, the audit-log gets spammed, and the cron-exit validator (`cron-exit-validator.js`) flags the run as needing attention. Operators eventually ignore the report and lose visibility into real failures.

## Reproduction

```bash
openclaw hybrid-mem maintenance nightly --verbose
# ... 25 minutes pass ...
Maintenance nightly — 36 steps (0 ok, 33 skipped, 0 deferred, 3 failed)
  extract-daily                skipped_guard     0.0s   guard (ran 14h ago)
  distill                      skipped_guard     0.0s   guard (ran 14h ago)
  resolve-contradictions       skipped_guard     0.0s   guard (ran 6h ago)
  enrich-entities              skipped_guard     0.0s   guard (ran 6h ago)
  extract-implicit             failed         1200.0s   extract-implicit partial failure (maxWallClock): 128 signals (106+/22-) from 116/471 sessions semantic=partial
  entity-mentions-cleanup      skipped_guard     0.0s   guard (ran 6h ago)
  dream-cycle-core             skipped_guard     0.0s   guard (ran 6h ago)
  continuous-verification      failed            0.7s   continuous-verification degraded (errors_present): checked=45 confirmed=0 stale=0 uncertain=45 errors=42 semantic=partial Machine status: status=degraded reason=errors_present checked=45 confirmed=0 stale=0 uncertain=45 errors=42
  ...
  audit-health                 failed            1.2s   6 warning(s) present and --strict was set
```

Note: on this installation the count was `7 warning(s)` (verified via `audit-health --strict --json`), but the orchestrator ran with `--strict` and emitted the same failure line.

### Failure 1: `extract-implicit` — `maxWallClock` (config `2026.6.256`)

**Code path:** `cli/cmd-feedback.js:447-595`

```js
// line 447
const maxWallClockSeconds = implicitCfg.maxWallClockSeconds ?? 300;
// line 467
const wallClockLimitReached = () => maxWallClockSeconds > 0 && (Date.now() - startTimeMs) / 1e3 >= maxWallClockSeconds;
// line 554
if (wallClockLimitReached()) {
  markPartialProgress("maxWallClock", deferredIncludingCurrentSession());
  break;
}
```

**Behaviour:** `extract-implicit` is **designed** to gracefully mark partial progress when it runs out of time budget. The `markPartialProgress("maxWallClock", ...)` flag is set, signals processed so far are inserted into `implicit_signals`, the loop breaks. The 128 signals extracted are real and persist.

**What goes wrong:** the orchestrator's step-result contract treats `partial=true` as `failed`. There is no distinction between:
- partial run that produced useful output (the `maxWallClock` case)
- partial run that errored mid-loop (a true failure)

Both get `semantic=partial` and both reach the cron-exit validator as failures.

**Why this matters:** the implicit-feedback step is **expected** to take longer than 20 minutes on a backlog this size (471 JSONL files, with the default 50-file-per-run guard). The 1200s config was set knowing the step would hit it; the partial behaviour is the *intended* graceful degradation. Yet the run still appears as a failure.

**Evidence this is the right behaviour:** the orchestrator reports `128 signals (106+/22-)` — that's the partial result. If we cared about completing all 471 sessions, we'd want either:
- A longer wall-clock budget (1200 → 1800+)
- A `--strict` flag on the orchestrator that treats `partial=true` as `failed`, defaulting to `partial=true → degraded-but-acceptable`
- A re-entry pattern where the next nightly resumes from `progress.sessionsVisited`

### Failure 2: `continuous-verification` — `errors_present`

**Code path:** `services/continuous-verifier.js:144-160` and `services/maintenance-orchestrator.js:160-164`

```js
// maintenance-orchestrator.js line 160
{
  name: "continuous-verification",
  tier: "nightly",
  guardIntervalMs: MAINTENANCE_GUARD_INTERVALS.h44,
  llmTier: "local",
  featureGate: (cfg) => cfg.verification?.enabled === true && cfg.verification?.continuousVerification === true
},
```

**Two-part root cause:**

**(a) Feature gate is broken.** `cfg.verification?.enabled === true && cfg.verification?.continuousVerification === true` — on this host `verification: {}` is unset, so the gate should be **false**, and the step should not run. But it ran. Either:
- The gate is bypassed somewhere (e.g. when called from `weekly-audit-health`'s message body)
- The default config resolves `cfg.verification?.continuousVerification` to truthy
- The orchestration always runs it regardless

**(b) `errors_present` is misclassified.** Even when the step does run, the verifier flags `UNCERTAIN` outcomes as `errors` (`continuous-verifier.js:158`), and the step reports `errors_present`. But `UNCERTAIN` is a valid verifier outcome (defined in the LLM prompt at line 16): the verifier legitimately returns `UNCERTAIN` when "no recent information available." Treating `UNCERTAIN` as an error means a normal "we have no recent context for this fact" outcome becomes a step failure.

**Evidence:** this run checked 45 facts, returned 0 `CONFIRMED`, 0 `STALE`, 45 `UNCERTAIN`, 42 errors. **All 45 verdicts were `UNCERTAIN`** — the verifier couldn't find recent context to confirm them. This is the intended behaviour for facts whose entities have no recent entries in `getRecentFacts(RECENT_FACTS_DAYS)`. The 42 errors are the LLM call timeouts that the verifier correctly caught and downgraded to `UNCERTAIN`.

### Failure 3: `audit-health --strict` on 7 advisory warnings

**Code path:** `cli/commands/manage/maintenance-step-runners.js:523-533` and `services/audit-health-exit-info.js:7-30`

```js
// maintenance-step-runners.js line 523
set("audit-health", async () => {
  const report = buildAuditHealthReport(b.factsDb, b.getMemoryCategories, b.cfg.entityExtraction?.stopWords ?? [], b.cfg.graph?.hubDegreeCap);
  const exitInfo = buildAuditHealthExitInfo({
    strict: true,  // <-- always true for the orchestrator
    warningCount: report.warningCount,
    errorCount: report.errorCount,
    ok: report.ok,
    status: report.status
  });
  if (exitInfo.exitCode !== 0) throw new Error(exitInfo.strictFailureReason ?? `${exitInfo.errorCount} error(s), ${exitInfo.warningCount} warning(s)`);
  return `warnings=${report.warningCount} errors=${report.errorCount}`;
});
```

```js
// audit-health-exit-info.js line 7
if (strict && (warningCount > 0 || errorCount > 0 || looksDegraded)) {
  const strictFailureReason = (() => {
    if (errorCount > 0 && warningCount > 0) return `${errorCount} error(s) and ${warningCount} warning(s) present and --strict was set`;
    if (errorCount > 0) return `${errorCount} error(s) present and --strict was set`;
    return `${warningCount} warning(s) present and --strict was set`;
  })();
  ...
}
```

**Root cause:** the orchestrator hard-codes `strict: true` for the audit-health step. There is no way to run the step non-strictly via the orchestrator. Yet the audit-health report's "warnings" category contains **advisory findings**, not errors:

| # | Warning | Severity | Should fail --strict? |
|---|---|---|---|
| 1 | `personaProposals scopeFilter not set` | config advisory | No |
| 2 | `Unconfigured categories meta=3` | migration drift | No |
| 3 | `Implicit-feedback patterns share a prefix 49×` | content-shape advisory | No |
| 4 | `Top entities include stop-word-like labels` | advisory | No |
| 5 | `959 active non-kv facts missing canonical embeddings` | size-of-gap report | No |
| 6 | `Vector lifecycle SLO breach (3.64% actual vs 2.00% target)` | SLO breach — *could* be configurable to fail | Maybe |
| 7 | `514 validated procedures not promoted` | awaiting-triage advisory | No |

The audit-health design clearly intends warnings to be informational ("here's what we found") but the step-runner hard-codes `strict: true` and fails on any non-zero warning count. The standalone CLI (`openclaw hybrid-mem audit-health`) defaults to non-strict; only `--strict` is strict. The orchestrator inverts this.

**Consequence:** every nightly run that surfaces a non-trivial fact store will fail. The 7 warnings above are *normal* for an installation of this size (60K+ facts, 14K+ distill facts, 17K+ recent facts loaded).

## Suggested fix paths

**For Failure 1 (`extract-implicit` partial):**

Minimum diff: in `services/maintenance-orchestrator.js`, the step-result contract should distinguish `partial=true, partialReason=maxWallClock` from `partial=true, partialReason=error`. The first should be `semantic=partial` *but not failure*; the second remains a failure.

Alternatively: in `services/maintenance-step-runners.js` (the step that runs `extract-implicit`), on `maxWallClock` partial, return `semantic=partial-success` rather than throwing. The `markPartialProgress` path produces useful output; let it succeed as a partial-success.

**For Failure 2 (`continuous-verification` degraded):**

Two options:

(a) **Fix the feature gate.** The orchestrator's gate `cfg.verification?.enabled === true && cfg.verification?.continuousVerification === true` should default to `enabled` for installations where `verification: {}` is the unset default — i.e. if `verification.enabled` is undefined and the verifier service is initialized, run it. Otherwise, skip cleanly.

(b) **Don't classify `UNCERTAIN` as an error.** In `continuous-verifier.js:158`, the `outcome = "UNCERTAIN"` after `result.errors++` should NOT increment `errors`. The verifier's job is to produce one of three verdicts; `UNCERTAIN` is a verdict, not an error.

(c) **Distinguish `partial` from `failed`** in the same way as Failure 1: a verifier run that produced 45 verdicts (all UNCERTAIN) is *complete*, not partial. The orchestrator should treat `checked > 0` as a successful run, regardless of outcome mix.

**For Failure 3 (`audit-health --strict` always-on):**

Minimum diff: in `cli/commands/manage/maintenance-step-runners.js:526`, change `strict: true` to `strict: false`. The `--strict` flag is meant to be opt-in, used when the operator wants to fail on warnings (e.g. for CI gates). The orchestrator's nightly run shouldn't fail the whole pipeline because of a 3.64% vectorless ratio.

Better: add a per-warning `failOnStrict` flag to `audit-health-json.js` so warnings 6 (`Vector lifecycle SLO breach`) could fail strictly while warnings 1–5 and 7 don't. The current all-or-nothing strict mode is too coarse.

## Impact

- Every nightly run reports `failed` exit code, triggering cron-exit validator alert paths
- `cron-exit-validator.js:697-699` flags the run as needing attention in the audit-log
- Operators learn to ignore the report; real failures get buried in the noise
- The `weekly-audit-health` cron (which is the *intended* place for strict audit) inherits the same always-on-strict behaviour from the step runner

## Environment

- Plugin: `openclaw-hybrid-memory@2026.6.256`
- OpenClaw: `2026.6.8 (844f405)`
- facts.db: 64,120+ facts; 14,638 distillation facts; 471 session JSONL files in `extract-implicit` backlog
- `cfg.verification`: `{}` (unset; the feature gate should be `false`, but the step ran)
- `audit-health` JSON report: 7 warnings, 0 errors, status `degraded`, ok `false`
- Run completed in ~25 minutes; the 3 "failures" are the only non-skipped steps
- Continuous-verification: 45 facts due, 45 UNCERTAIN verdicts, 42 LLM call errors downgraded to UNCERTAIN

## Related work

- #1942 (closed, shipped 2026.6.250): orchestrator abort on degraded backlog. The previous fix prevented the orchestrator from aborting on degraded state — but it didn't address the fact that the orchestrator still treats `partial` and `degraded` as failure for cron-exit purposes.
- Cron-exit validator (`services/cron-exit-validator.js:697-699`): classifies continuous-verification failure when `Machine status: status=degraded` is detected. This validator is downstream of the orchestrator; the upstream fix should remove the failure condition, not the validator.

### LINKED ISSUE #1950: memory_store tool: 'Cannot read properties of undefined (reading trim)' when model emits truncated/empty args
## Summary

Calling `memory_store` via the agent tool dispatcher throws `Cannot read properties of undefined (reading 'trim')` at `dist/tools/memory/register-store-tools.js:83` when the model's tool-call args fail to parse. The OpenClaw gateway then logs:

```
[tools] memory_store failed: Cannot read properties of undefined (reading 'trim') raw_params={"id":"memory_store","args":{}}
```

The `args` is `{}` — not `undefined` — but the schema-defined `text` field is missing because the OpenClaw streaming JSON parser silently returned an empty object instead of either passing through the raw args or producing a typed error.

This is reproducible from the agent tool surface (when the model streams incomplete JSON args for `memory_store`). The CLI form (`openclaw hybrid-mem memory store --text "..."`) works correctly because it bypasses the streaming JSON layer entirely.

## Root cause chain

The failure involves three layers. All three are needed to produce the symptom:

### Layer 1 (gateway): silent empty-object fallback in streaming JSON parser

`/mnt/openclaw/node_modules/openclaw/dist/json-parse-CydVzlvP.js:115-138` — `parseStreamingJson(partialJson)`:

```js
function parseStreamingJson(partialJson) {
    if (!partialJson || partialJson.trim() === "") return {};
    try {
        return asStreamingJsonRecord(parseJsonWithRepair(partialJson));
    } catch {
        try {
            return asStreamingJsonRecord(parse(partialJson));
        } catch {
            try {
                return asStreamingJsonRecord(parse(repairJson(partialJson)));
            } catch {
                return {};
            }
        }
    }
}
```

Three parse strategies (full JSON parse + repair → partial-json lib → partial-json lib + repair). If all three fail, the function returns `{}` **silently** — no error, no logging, no telemetry. The caller cannot distinguish between "the model emitted no args" and "the model's args failed to parse".

**Where the empty result gets used:** `/mnt/openclaw/node_modules/openclaw/dist/openai-completions-DCEsPPkI.js:439` and `:588`:

```js
block.arguments = parseStreamingJson(block.partialArgs);
```

This runs every time the model streams a tool-call delta. If the model emits something like `{"te` (truncated, before completing `text`), the streaming parser attempts three parses, all fail, and `block.arguments` becomes `{}`.

### Layer 2 (gateway): dispatcher does not validate emptiness

`/mnt/openclaw/node_modules/openclaw/dist/tool-split-DXB3RqC6.js:235-280` — the dispatcher:

```js
} catch (err) {
    if (signal?.aborted) throw err;
    if (isBeforeToolCallBlockedError(err)) { ... }
    const described = describeToolExecutionError(err);
    if (described.stack && described.stack !== described.message) logDebug(`tools: ${normalizedName} failed stack:\n${described.stack}`);
    const inputPreview = describeToolFailureInputs({
        toolName: normalizedName,
        rawParams: params,
        effectiveParams: executeParams
    });
    logError(`[tools] ${normalizedName} failed: ${described.message} ${inputPreview}`);
    return buildToolExecutionErrorResult({
        toolName: normalizedName,
        message: described.message
    });
}
```

The dispatcher propagates the exception to the framework but does not pre-check whether `params` is empty or missing required fields. The `validateToolArguments` function (`src-M7TBQdDX.js:258`) **does** validate the args against the tool's TypeBox schema — but only if `toolCall.arguments` is an object. Empty `{}` will fail validation because `text` is required (`Type.String({ description: "Information to remember" })`). However, `validateToolArguments` throws a validation error message like:

```
Validation failed for tool "memory_store":
  - /: must have required property 'text'
  ...
```

It does NOT throw `Cannot read properties of undefined (reading 'trim')` — the trim error comes from the plugin handler itself.

### Layer 3 (plugin): the throwing line

`dist/tools/memory/register-store-tools.js:83`:

```js
async execute(_toolCallId, params) {
    try {
        const { text, why, importance = .5, category = "other", ... } = params;
        const { factsDb: storeFactsDb, vectorDb: storeVectorDb } = resolveToolVaultBackends(runtime, typeof vaultParam === "string" ? vaultParam : void 0);
        ...
        if (text.trim().length === 0) return {  // ← line 83, throws on undefined
            content: [{ type: "text", text: "memory_store: text is required and must be non-empty" }]
        };
```

When `params = {}`, the destructure leaves `text = undefined`. `text.trim()` throws `TypeError: Cannot read properties of undefined (reading 'trim')`.

**Compare with the fix already used elsewhere in the same plugin:** `dist/tools/memory/register-agent-verb-tools.js:215` correctly uses `String(args.text ?? "").trim()`. The `memory_store` handler should use the same defensive pattern.

### From the agent tool surface (the failure case)

1. Use a session where the model is configured to call `memory_store` but the JSON args stream is interrupted (network drop, model timeout, partial-output truncation). On this installation, the model is `minimax/MiniMax-M3`. The model emits `{"te` (truncated).
2. The gateway log shows:
   ```
   [tools] memory_store failed: Cannot read properties of undefined (reading 'trim') raw_params={"id":"memory_store","args":{}}
   ```
3. The tool returns an error result to the model, which then typically retries, but the retry may also fail for the same reason.

### From the CLI (does not reproduce)

```bash
$ openclaw hybrid-mem memory store --text "test"
  ok: stored fact #12345
```

The CLI uses a different code path that bypasses the streaming JSON parser. This is why CLI usage appears healthy while agent tool usage fails.

## Suggested fix paths

There are **three** candidate fixes, in order of intrusiveness:

### Option A: Defensive pattern in the plugin (minimum diff, recommended for plugin repo)

Replace `text.trim()` with `String(text ?? "").trim()` in `register-store-tools.js:83`. Same pattern as `register-agent-verb-tools.js:215`. Plus add a clearer error message when `text` is missing so future debugging is easier.

```js
// dist/tools/memory/register-store-tools.js line 83
const trimmedText = String(text ?? "").trim();
if (trimmedText.length === 0) return {
    content: [{
        type: "text",
        text: "memory_store: text is required and must be non-empty (received empty or non-string text)"
    }]
};
```

This is a 2-line fix and matches the existing pattern. **Apply in `openclaw-hybrid-memory` repo.**

### Option B: Surface streaming JSON parse failures in the gateway

In `dist/json-parse-CydVzlvP.js:138`, instead of returning `{}` silently when all three parse strategies fail, emit a diagnostic event (`emitTrustedDiagnosticEvent` is already imported) and return a sentinel that the dispatcher can recognise. The dispatcher then translates the sentinel into a tool-level error like "tool arguments could not be parsed from model output; please retry."

This belongs in the **`openclaw` core repo**, not the plugin. **Upstream.**

### Option C: Schema validation before dispatch

In `dist/tool-split-DXB3RqC6.js` (the dispatcher), call `validateToolArguments(tool, { name, arguments: params })` before invoking `tool.execute`. The existing `validateToolArguments` (in `src-M7TBQdDX.js:258`) already produces a clear error for `text` missing:

```
Validation failed for tool "memory_store":
  - /: must have required property 'text'
```

This is much more useful than the cryptic `.trim()` error. The dispatcher should fail-fast with a schema-validated error rather than letting the tool handler throw. **Upstream.**

### Recommendation

**Apply Option A immediately in the plugin repo** as a defensive fix — it's tiny and matches existing pattern. **File Options B and C as separate upstream issues in `openclaw` core repo** because the root cause is the silent empty-object fallback in the streaming JSON parser, not the plugin.

## Environment

- Plugin: `openclaw-hybrid-memory@2026.6.256`
- OpenClaw core: `2026.6.8 (844f405)`
- Model: `minimax/MiniMax-M3` (inherits OpenAI streaming semantics via `openai-completions-DCEsPPkI.js`)
- Observed via: agent tool dispatcher in Telegram session
- CLI form works correctly; only the agent tool surface fails

## Related work

- Tool-call streaming JSON parser: `/mnt/openclaw/node_modules/openclaw/dist/json-parse-CydVzlvP.js` (silent empty-object fallback at line 138)
- Tool dispatcher: `/mnt/openclaw/node_modules/openclaw/dist/tool-split-DXB3RqC6.js:235-280` (no pre-dispatch schema validation)
- Existing defensive pattern in the same plugin: `dist/tools/memory/register-agent-verb-tools.js:215` (`String(args.text ?? "").trim()`)
- AGENTS.md error pattern: `dist/cli/commands/manage/maintenance-step-runners.js` already uses defensive `?? ""` patterns for tool params

## Diagnostic note

The `[tools] memory_store failed: Cannot read properties of undefined (reading 'trim') raw_params={"id":"memory_store","args":{}}` log line is the exact symptom. When the operator sees this in the gateway log, the immediate workaround is to retry (the model's next attempt usually produces a complete JSON stream). When retries also fail, the cause is upstream — the streaming parser is being given malformed input from the model layer, not from user input.

markus-lassfolk:
## Summary

Combined fix for two unrelated false-failure / crash paths:

### #1949 — Maintenance nightly misleading exit codes

- **extract-implicit:** Budget-cap partial runs (`maxWallClock`, session/signal/trajectory caps) succeed with `semantic=monitoring` instead of failing the pipeline.
- **continuous-verification:** LLM downgrades to `UNCERTAIN` no longer increment `errors`; all-uncertain completed cycles are healthy. Cron validator ignores legacy `all_uncertain` degraded machine-status lines.
- **audit-health:** Orchestrator step runs non-strict (warnings only); report errors still fail the step.
- **Guard advancement:** Monitoring-only degraded summary semantics no longer block cron guard updates when there are no reportable blocking issues (`360597ba1`).

### #1950 — memory_store crash on empty tool args

- Require string `text` before validation so `{}` or malformed streaming args return structured `invalid_text` instead of throwing on `undefined.trim()` or coercing objects to `[object Object]`.

Closes #1949. Closes #1950.

## Test plan

- [x] Targeted: `dream-cycle-followup-logging`, `continuous-verifier`, `cron-exit-validator`, `memory-store-early-validation`, `maintenance-validator-coverage`
- [x] Codex P1/P2 review feedback addressed in `360597ba1`
- [ ] Host: `maintenance nightly --verbose` exits 0 for prior false-failure scenario
- [ ] Host: agent `memory_store` with truncated args returns `invalid_text`, not TypeError

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ---
 
+## [2026.6.257] - 2026-06-25
+
+### Fixed
+
+- **Maintenance nightly exit codes (#1949):** Treat `extract-implicit` budget-cap partial runs (`maxWallClock`, session/signal/trajectory caps) as successful monitoring steps instead of failures; stop counting continuous-verification LLM downgrades to `UNCERTAIN` as errors and do not fail nightly when all facts receive uncertain verdicts; run orchestrator `audit-health` non-strict (warnings only) while still failing on report errors. Cron exit validator ignores legacy `all_uncertain` degraded machine-status lines.
+
+---
+
 ## [2026.6.256] - 2026-06-25
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ### Fixed
 
+<<<<<<< HEAD
 - **Maintenance nightly exit codes (#1949):** Treat `extract-implicit` budget-cap partial runs (`maxWallClock`, session/signal/trajectory caps) as successful monitoring steps instead of failures; stop counting continuous-verification LLM downgrades to `UNCERTAIN` as errors and do not fail nightly when all facts receive uncertain verdicts; run orchestrator `audit-health` non-strict (warnings only) while still failing on report errors. Cron exit validator ignores legacy `all_uncertain` degraded machine-status lines.
+=======
+- **memory_store empty args (#1950):** Guard missing or non-string `text` with `String(text ?? "").trim()` so streaming-parser `{}` args return a structured `invalid_text` error instead of throwing `Cannot read properties of undefined (reading 'trim')`.
+>>>>>>> 2b1d5c0ec (fix(memory_store): handle missing text from empty tool args (#1950))
 
 ---
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,11 +31,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ### Fixed
 
-<<<<<<< HEAD
 - **Maintenance nightly exit codes (#1949):** Treat `extract-implicit` budget-cap partial runs (`maxWallClock`, session/signal/trajectory caps) as successful monitoring steps instead of failures; stop counting continuous-verification LLM downgrades to `UNCERTAIN` as errors and do not fail nightly when all facts receive uncertain verdicts; run orchestrator `audit-health` non-strict (warnings only) while still failing on report errors. Cron exit validator ignores legacy `all_uncertain` degraded machine-status lines.
-=======
-- **memory_store empty args (#1950):** Guard missing or non-string `text` with `String(text ?? "").trim()` so streaming-parser `{}` args return a structured `invalid_text` error instead of throwing `Cannot read properties of undefined (reading 'trim')`.
->>>>>>> 2b1d5c0ec (fix(memory_store): handle missing text from empty tool args (#1950))
+- **memory_store empty args (#1950):** Require string `text` before validation so streaming-parser `{}` or malformed args return structured `invalid_text` instead of throwing on `undefined.trim()` or coercing objects to `[object Object]`.
+- **Maintenance guard (#1949 review):** Allow cron guard advancement when summary validation reports monitoring-only degraded semantics with no blocking issues.
 
 ---
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ---
 
-## [2026.6.257] - 2026-06-25
+## [2026.6.258] - 2026-06-25
 
 ### Fixed
 
--- a/extensions/memory-hybrid/cli/commands/manage/dream-cycle-followup-steps.ts
+++ b/extensions/memory-hybrid/cli/commands/manage/dream-cycle-followup-steps.ts
@@ -9,7 +9,9 @@ import { runCrystallizationProposalCycle } from "../../../services/crystallizati
 import { getEffectivenessReport, runClosedLoopAnalysis } from "../../../services/feedback-effectiveness.js";
 import {
   assessContinuousVerificationResult,
+  extractImplicitSemanticOutcome,
   formatContinuousVerificationAssessmentLine,
+  isGracefulExtractImplicitPartial,
 } from "./dream-cycle-followup.js";
 
 export interface DreamCycleFollowUpDeps {
@@ -66,8 +68,8 @@ export async function runExtractImplicitStep(deps: DreamCycleFollowUpDeps, verbo
     includeClosedLoop: false,
     verbose,
   });
-  const summary = `${res.signalsExtracted} signals (${res.positiveCount}+/${res.negativeCount}-) from ${res.sessionsProcessed}/${res.sessionsScanned} sessions semantic=${res.partial ? "partial" : "success"}`;
-  if (res.partial) {
+  const summary = `${res.signalsExtracted} signals (${res.positiveCount}+/${res.negativeCount}-) from ${res.sessionsProcessed}/${res.sessionsScanned} sessions semantic=${extractImplicitSemanticOutcome(res)}`;
+  if (res.partial && !isGracefulExtractImplicitPartial(res.partialReason)) {
     throw new Error(`extract-implicit partial failure (${res.partialReason ?? "capped"}): ${summary}`);
   }
   return summary;
--- a/extensions/memory-hybrid/cli/commands/manage/dream-cycle-followup.ts
+++ b/extensions/memory-hybrid/cli/commands/manage/dream-cycle-followup.ts
@@ -20,6 +20,24 @@ export interface ContinuousVerificationAssessment {
   summary?: string;
 }
 
+/** Budget-cap partial runs that persist useful work and resume on the next nightly. */
+const GRACEFUL_EXTRACT_IMPLICIT_PARTIAL_REASONS = new Set([
+  "maxWallClock",
+  "maxSessions",
+  "maxSignals",
+  "maxTrajectories",
+]);
+
+export function isGracefulExtractImplicitPartial(partialReason?: string): boolean {
+  return partialReason != null && GRACEFUL_EXTRACT_IMPLICIT_PARTIAL_REASONS.has(partialReason);
+}
+
+export function extractImplicitSemanticOutcome(res: { partial?: boolean; partialReason?: string }): string {
+  if (!res.partial) return "success";
+  if (isGracefulExtractImplicitPartial(res.partialReason)) return "monitoring";
+  return "partial";
+}
+
 export function formatExtractImplicitFeedbackProgress(
   snapshot: ExtractImplicitFeedbackProgressSnapshot | undefined,
 ): string | undefined {
@@ -57,24 +75,17 @@ export function formatExtractImplicitFeedbackProgress(
 }
 
 export function assessContinuousVerificationResult(result: VerificationCycleResult): ContinuousVerificationAssessment {
-  if (result.errors > 0) {
+  const verdicts = result.confirmed + result.stale + result.uncertain;
+  if (result.errors > 0 && (result.checked === 0 || verdicts < result.checked)) {
     const summary =
       result.checked === 0
         ? `infrastructure error prevented verification (${result.errors} error(s))`
-        : `${result.errors}/${result.checked} verification check(s) errored`;
+        : `${result.errors}/${result.checked} verification check(s) failed without a verdict`;
     return {
       status: "degraded",
       reason: "errors_present",
       shouldFailPipeline: true,
       summary,
-    };
-  }
-  if (result.checked > 0 && result.confirmed === 0 && result.stale === 0) {
-    return {
-      status: "degraded",
-      reason: "all_uncertain",
-      shouldFailPipeline: true,
-      summary: `all ${result.checked} verification check(s) were uncertain`,
     };
   }
   return { status: "healthy", reason: "healthy", shouldFailPipeline: false };
--- a/extensions/memory-hybrid/cli/commands/manage/maintenance-step-runners.ts
+++ b/extensions/memory-hybrid/cli/commands/manage/maintenance-step-runners.ts
@@ -19,7 +19,6 @@ import {
 import { syncLifecycleFromGitHub } from "../../../services/lifecycle/github-adapter.js";
 import { runAnalyzeMaintenanceLogs } from "./register-analyze-maintenance-logs.js";
 import { buildAuditHealthReport } from "./storage-stats-helpers.js";
-import { buildAuditHealthExitInfo } from "../../../services/audit-health-exit-info.js";
 import { recordStorageGrowthSample } from "./storage-stats-helpers.js";
 import { runPendingDigestAutopilotCron } from "../../../services/pending-digest-autopilot-cron.js";
 import { buildPendingReviewDigestReport } from "../../../services/pending-review-digest.js";
@@ -40,6 +39,7 @@ import {
   runExtractImplicitStep,
   runToolEffectivenessStep,
 } from "./dream-cycle-followup-steps.js";
+import { extractImplicitSemanticOutcome, isGracefulExtractImplicitPartial } from "./dream-cycle-followup.js";
 import { cleanupImplicitFeedbackDuplicates } from "../../cmd-feedback.js";
 import { resolveScanMaintenanceOverrides, type ScanMaintenanceOverrideInput } from "../../maintenance-overrides.js";
 import { nowIso } from "../../../utils/dates.js";
@@ -142,12 +142,13 @@ function formatExtractImplicitSummary(res: {
   sessionsProcessed: number;
   sessionsScanned: number;
   partial?: boolean;
+  partialReason?: string;
 }): string {
-  return `${res.signalsExtracted} signals (${res.positiveCount}+/${res.negativeCount}-) from ${res.sessionsProcessed}/${res.sessionsScanned} sessions semantic=${res.partial ? "partial" : "success"}`;
+  return `${res.signalsExtracted} signals (${res.positiveCount}+/${res.negativeCount}-) from ${res.sessionsProcessed}/${res.sessionsScanned} sessions semantic=${extractImplicitSemanticOutcome(res)}`;
 }
 
 function assertExtractImplicitNotPartial(res: { partial?: boolean; partialReason?: string }, summary: string): void {
-  if (res.partial) {
+  if (res.partial && !isGracefulExtractImplicitPartial(res.partialReason)) {
     throw new Error(`extract-implicit partial failure (${res.partialReason ?? "capped"}): ${summary}`);
   }
 }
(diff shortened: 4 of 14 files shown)


markus-lassfolk:
## Review triage (commit `360597ba1`)

**Codex P1 — monitoring summaries must advance the guard:** Addressed in `shouldUpdateMaintenanceGuard()` — when `semanticStatus=degraded` but there are no reportable blocking issues (summary-only `semantic=monitoring` from budget-cap `extract-implicit`, advisory `audit-health`, etc.), guard advancement is now allowed. Covered by `validateFromSummaryJson` test: *updates guard for successful nightly summary with monitoring-only step semantics*.

**Codex P2 (from superseded #1952) — reject non-string `text`:** Addressed in `register-store-tools.ts` via `typeof text === "string"` before trim; regression test added for object-typed `text`.

**Bugbot:** Skipped (usage limit) — no findings to triage.

**Forge:** Dispatched on earlier SHA; no new actionable feedback after `360597ba1`.

